PR #10, opened 2015-11-13
Separate readFile promise for better error handling

TrySound:


TrySound:
Do not merge. Want to make it with all loaders. Waiting for success

davidtheclark:
Ah, this technique makes sense to me. I had some struggles figuring out the right way to handle the interface between errors happening in the `readFile` callback and the Promise chain (http://stackoverflow.com/questions/33651306/why-doesnt-promise-catch-error-thrown-by-async-operation). I like the way you are doing it here.

TrySound:
Will make some noise without local tests. )

TrySound:
@davidtheclark  I see eslint config is not quite strict

TrySound:
Seems done. Should I squash all commits?

davidtheclark:
Nah, I haven't been too worried about a clean history during this initial development period.

PR #14, opened 2015-11-15
Index refactoring

TrySound:
Everything is the same, but shorter :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,7 +10,6 @@ var loadRc = require('./lib/loadRc');
 var loadJs = require('./lib/loadJs');
 var loadDefinedFile = require('./lib/loadDefinedFile');
 
-var DONE = 'done';
 var parsedCliArgs = minimist(process.argv);
 
 module.exports = function(moduleName, options) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -29,75 +28,35 @@ module.exports = function(moduleName, options) {
...
+    var currentSearchPath = joinPath(splitSearchPath);
+
+    return Promise.resolve(null).then(function() {
+      if (!options.packageProp) return;
+      return loadPackageProp(currentSearchPath, options.packageProp);
+    }).then(function(result) {
+      if (result || !options.rc) return result;
+      return loadRc(path.join(currentSearchPath, options.rc), {
+        strictJson: options.rcStrictJson,
+      });
```
davidtheclark: Oooo, this is cool: taking advantage of the fact that you can return a synchronous value *or* a Promise and they are treated the same way. Working on this module has taught me a lot about Promises :)

davidtheclark:
Nice job, thanks!

index.js
TrySound: Yeah, they are very cool! :)

davidtheclark:
@TrySound : How are you feeling about the API now?

TrySound:
Seems much better

index.js
jeddy3: > Working on this module has taught me a lot about Promises :)

Likewise, I've learnt a lot from just seeing the evolution of the code both here and in stylelint's `standalone.js`. So, thanks for that :)

TrySound: My pleasure :)

PR #16, opened 2015-11-16
Add filenames in parsers for better error message

TrySound:


davidtheclark:
Good addition, thanks.

PR #17, opened 2015-11-17
Switch from tape to ava

davidtheclark:
This was primarily for the sake of experimenting with [ava](https://github.com/sindresorhus/ava) -- but I think it's actually better, because the Promise support of ava leads to cleaner tests for this Promise-heavy project (no need to count assertions). @TrySound what do you think?

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@
 
 var path = require('path');
 var oshomedir = require('os-homedir');
-var Promise = require('pinkie-promise');
+var Promise = require('bluebird');
 var minimist = require('minimist');
 var assign = require('object-assign');
 var loadPackageProp = require('./lib/loadPackageProp');
--- a/lib/loadDefinedFile.js
+++ b/lib/loadDefinedFile.js
@@ -1,7 +1,7 @@
 'use strict';
 
 var fs = require('graceful-fs');
-var Promise = require('pinkie-promise');
+var Promise = require('bluebird');
 var yaml = require('js-yaml');
 var parseJson = require('parse-json');
 var requireFromString = require('require-from-string');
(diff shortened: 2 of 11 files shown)


davidtheclark:
I have no idea why only Node 0.10 is complaining about a bunch of `Unhandled Rejection`s --- when they are in fact handled by `.catch()` statements.

TrySound:
@davidtheclark Why they are serial?

TrySound:
@davidtheclark Success :)

davidtheclark:
@TrySound They had to be serial because of the way that I'm stubbing `fs` each time. If I don't mark them as serial, one starts and stubs `fs`, then another starts and tries to stub what's already stubbed. I'm open to another way of stubbing/mocking `fs`; but this way works for now.

TrySound:
@davidtheclark Okay

TrySound:
@davidtheclark Hey, where's ci?

davidtheclark:
@TrySound ?? Looks like it passed

TrySound:
@davidtheclark Yep. Just was big delay. But pinkie is buggy :(

davidtheclark:
@TrySound So I don't have any other plans currently. Do you? Or are we ready for a 1.0.0 release?

TrySound:
@davidtheclark I want to try to understand why pinkie doesn't work. It's just on my computer and appveyor

TrySound:
@davidtheclark Okay, seems like it's windows hello.

TrySound:
@davidtheclark I'm not ready for new plans :)) We can make patches if something will go wrong.

davidtheclark:
Ok. Yeah, we might have bugs but I think the API looks good. Sometime later today or tomorrow I will take another look at the docs and then release.

MoOx:
\o/

davidtheclark:
Released!

TrySound:
Cool!

PR #23, opened 2016-01-10
Add rcExtensions option; closes #18

### LINKED ISSUE #18: Support for rc files with extensions
I read about the latest ESLint release (http://eslint.org/blog/2015/11/eslint-v1.10.0-released/) and thought that what they're doing with rc file extensions would be worth mimicking.

The idea is that rc files would not have to be extensionless: if user adds an extension to the file (e.g. `.stylelintrc.yaml`), then they'll get syntax highlighting in their editor and more exact error stacks if there are syntax errors.

davidtheclark:
This adds an `rcExtensions` option, which causes cosmiconfig to look for rc files with the following extensions (in order): `json`, `yaml`, `yml`, `js`.

(Should we cut `yml`?)

One change I made that is not relevant to this particular functionality is to use `_.get()` in some tests, because I found I could get more helpful error messages that way, sometimes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Head
+
+- Add `rcExtensions` option.
+
 ## 1.0.2
 
 - Fix handling of `require()`'s within JS module configs.
--- a/README.md
+++ b/README.md
@@ -18,6 +18,9 @@ cosmiconfig continues to search in these places all the way down the file tree u
 
 Additionally, all of these search locations are configurable: you can customize filenames or turn off any location.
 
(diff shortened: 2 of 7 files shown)


davidtheclark:
@TrySound Does this look good to you?

davidtheclark:
@MoOx, @jeddy3 — any opinions? I just want to get some confirmation that this approach makes sense to people before I merge and publish.

TrySound:
@davidtheclark Sorry, skipped. Seems good to me.

jeddy3:
Likewise :)

davidtheclark:
Thanks! I'll merge tonight and push a new release. Any objections to adding this feature to stylelint?

jeddy3:
None. I think it'll make a great addition :)

PR #24, opened 2016-03-30
Prevent cosmiconfig from immediately monkey patching `fs`

m-allanson:
Prevent cosmiconfig from immediately monkey patching `fs`

This allows cosmiconfig to be imported in environments that don't have access to the `fs` module. It does not make it possible to run cosmiconfig in these environments.

When required, the `graceful-fs` module attempts to immediately monkey patch node's `fs` module. This (understandably) causes errors if running in an environment without an `fs` module.

This change means that `graceful-fs` will not run until cosmiconfig is invoked, instead of when cosmiconfig is required or imported by another module.

This should help with https://github.com/stylelint/stylelint/issues/993

### DIFF
--- a/lib/loadDefinedFile.js
+++ b/lib/loadDefinedFile.js
@@ -1,12 +1,12 @@
 'use strict';
 
-var fs = require('graceful-fs');
 var Promise = require('pinkie-promise');
 var yaml = require('js-yaml');
 var parseJson = require('parse-json');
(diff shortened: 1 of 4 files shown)


TrySound:
@m-allanson I think such cases should be handled with bundler like browserify. Unnecessary stuff will be just replaced with empty module.

m-allanson:
@TrySound Ok I'll give that a try.

Issue #28, opened 2016-10-09
Promise not run `resolve` or `reject` when config file not exist.


gucong3000:
Windows can reproduce the bug

```javascript
return stylelint.getConfigForFile("E:\\work\\stylefmt\\test\\fixtures\\at-apply\\at-apply.css")

.then(config => {
  // not run when configfile not exist.
  console.log(config)
})

.catch(ex => {
  // not run when configfile not exist.
  console.error(ex)
})
```

https://github.com/davidtheclark/cosmiconfig/blob/master/lib/createExplorer.js#L79
Always `false` in Windows. It is infinite loop.

davidtheclark:
@gucong3000 Does this bug only happen when you pass windows-style path strings? What if you pass unix-style path strings, or use Node's `path.join` to create the path?

davidtheclark:
Also, that's not a valid Windows path, right? Why the double-slashes? What happens if you make them single slashes — does it work?

gucong3000:
@davidtheclark 

Valid path in Windows can reproduce the bug.

> Does this bug only happen when you pass windows-style path strings?

Yes.

> What if you pass unix-style path strings, or use Node's path.join to create the path?

If not run in Windows, It is work properly.


> that's not a valid Windows path, right?

that's valid Windows path.

> Why the double-slashes?

Escape character. please try `console.log("E:\\")`, this will output single slash, like `E:\`

gucong3000:
```javascript
test.serial('Configuration file not exist', (assert) => {
  const loadConfig = cosmiconfig('not_exist_rc_name').load;
  return loadConfig('.').then((result) => {
    assert.deepEqual(result, null);
  });
});
```
Running this under Windows , the states of Promise will always be "pending""

gucong3000:
There is tow way can fix.

1. `splitPath()` only run once.
2. do not use `path.resolve()` in `splitPath()`

[Specific reasons for bug production](https://github.com/davidtheclark/cosmiconfig/pull/31#discussion_r82718656)

davidtheclark:
Closed by #29

PR #29, opened 2016-10-09
#28 fix infinite loop in windows

gucong3000:


davidtheclark:
Thanks @gucong3000! Is there a way to add a test to prevent against regressions?

I'll merge this to release ASAP, but please consider adding a test in another PR.

lib/createExplorer.js
```diff
@@ -47,9 +47,6 @@ module.exports = function (options) {
         const directory = (searchPathIsDirectory)
           ? absoluteSearchPath
           : path.dirname(absoluteSearchPath);
-        if (directoryCache && directoryCache.has(directory)) {
-          return directoryCache.get(directory);
-        }
```
davidtheclark: We cannot remove these lines. Using the directory cache is essential to prevent lots and lots of unnecessary file system activity.

Why is the cache causing this problem? What is the actual problem? And can you reproduce it in a test?

davidtheclark: Ok ... so this doesn't relate to the bug, but is just code cleanup — is that right?

PR #31, opened 2016-10-10
Fix infinite loop on Windows when configuration file not exist.

gucong3000:
#28

### DIFF
--- a/lib/createExplorer.js
+++ b/lib/createExplorer.js
@@ -71,9 +71,10 @@ module.exports = function (options) {
       })
       .then((result) => {
         if (result) return result;
-        if (!splitSearchPath) splitSearchPath = splitPath(directory);
-        if (!splitSearchPath.pop() || directory === options.stopDir) return null;
-        return searchDirectory(joinPath(splitSearchPath), splitSearchPath);
+        if (!splitSearchPath) splitSearchPath = directory.split(path.sep);
+        if (splitSearchPath.length <= 1 || directory === options.stopDir) return null;
+        splitSearchPath.pop();
+        return searchDirectory(splitSearchPath.join(path.sep), splitSearchPath);
       })
       .then(transform);
 
@@ -96,15 +97,6 @@ function isDirectory(filepath) {
       return resolve(stats.isDirectory());
     });
   });
-}
-
-function splitPath(x) {
-  return path.resolve(x || '').split(path.sep);
-}
-
-function joinPath(parts) {
-  if (!parts) return parts;
-  return parts.join(path.sep);
 }
 
 function identityPromise(x) {
--- a/lib/createExplorer.js
+++ b/lib/createExplorer.js
@@ -112,6 +112,7 @@ function splitPath(x) {
 }
 
 function joinPath(parts) {
+  if (!parts) return parts;
   return parts.join(path.sep);
 }
 
--- a/lib/createExplorer.js
+++ b/lib/createExplorer.js
@@ -47,20 +47,13 @@ module.exports = function (options) {
         const directory = (searchPathIsDirectory)
           ? absoluteSearchPath
           : path.dirname(absoluteSearchPath);
-        if (directoryCache && directoryCache.has(directory)) {
-          return directoryCache.get(directory);
-        }
-        return searchDirectory(splitPath(directory));
+        return getDirectoryCache(directory) || searchDirectory(splitPath(directory));
       });
   }
 
   function searchDirectory(splitSearchPath) {
     const directory = joinPath(splitSearchPath);
-    if (directoryCache && directoryCache.has(directory)) {
-      return directoryCache.get(directory);
-    }
-
-    const result = Promise.resolve()
+    const result = getDirectoryCache(directory) || Promise.resolve()
       .then(() => {
         if (!options.packageProp) return;
         return loadPackageProp(directory, options);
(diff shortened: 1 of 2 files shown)


lib/createExplorer.js
```diff
@@ -47,14 +47,11 @@ module.exports = function (options) {
         const directory = (searchPathIsDirectory)
           ? absoluteSearchPath
           : path.dirname(absoluteSearchPath);
-        if (directoryCache && directoryCache.has(directory)) {
```
gucong3000: Redundant. Same with [#L58](#diff-a2884610eeadaee7016bf1b9cd9a26d6L58).

lib/createExplorer.js
```diff
@@ -74,11 +71,9 @@ module.exports = function (options) {
       })
       .then((result) => {
         if (result) return result;
-
-        const nextDirectory = joinPath(splitPath(directory).slice(0, -1));
-        if (!nextDirectory || directory === options.stopDir) return null;
-
-        return searchDirectory(nextDirectory);
+        if (!splitSearchPath) splitSearchPath = splitPath(directory);
+        if (!splitSearchPath.pop() || directory === options.stopDir) return null;
```
davidtheclark: Can we not use `pop()` here? I don't like how it mutates the array at the same time that we're using it to make an assertion. I'd prefer accessing the item in the array without mutation, and creating a new array without it.

lib/createExplorer.js
```diff
@@ -74,11 +71,9 @@ module.exports = function (options) {
...
+        return searchDirectory(joinPath(splitSearchPath), splitSearchPath);
```
davidtheclark: For my own sake, would you mind trying to explain why these minor changes pix the bug? I'm really curious.

test/failed-directories.test.js
```diff
@@ -330,3 +330,12 @@ test.serial('with rcExtensions, find invalid JS in .foorc.js', (assert) => {
     assert.is(error.name, 'SyntaxError', 'threw correct error type');
   });
 });
+
+test.serial('Configuration file not exist', (assert) => {
+
+  const loadConfig = cosmiconfig('not_exist_rc_name').load;
+
+  return loadConfig('.').then((result) => {
+    assert.deepEqual(result, null);
```
davidtheclark: You can use `assert.is` here.

test/failed-directories.test.js
```diff
@@ -330,3 +330,12 @@ test.serial('with rcExtensions, find invalid JS in .foorc.js', (assert) => {
     assert.is(error.name, 'SyntaxError', 'threw correct error type');
   });
 });
+
+test.serial('Configuration file not exist', (assert) => {
```
davidtheclark: This new test is fine. Just to confirm: Does it fail on Windows without the above changes? It passes fine on my Mac, but that's to be expected.

gucong3000: I am sure, It fail  on Windows 10 without the above changes.The specific phenomenon is that the console, `ava` long time no response.

test/failed-directories.test.js
gucong3000: OK.

lib/createExplorer.js
gucong3000: In the process of finding the configuration file, each loop is required to split the string into an array, then merge array into a string.
I think in the first loop split strings into an array, after only merger this array, Is is more efficient.

davidtheclark: I do not like relying on a side-effect within a condition. It's "efficient" in the sense of reducing lines of code — but that's not the goal. It's much less clear to read, and when I come back to this a year from now and need to remind myself of what's going on, it will be easier to do so with the side-effect in a condition.

lib/createExplorer.js
gucong3000: here's the problem.
```javascript
function splitPath(x) {
  return path.resolve(x || '').split(path.sep);
}
```

I did some experiments on the command line.

```
E:\work\cosmiconfig
$ node -v
v6.7.0

E:\work\cosmiconfig
$ node
> path.resolve("C:")
'C:\\Users\\cdyf'
> path.resolve("D:")
'D:\\Program Files\\cmder\\config\\profile.d'
> path.resolve("E:")
'E:\\work\\cosmiconfig'
>
(To exit, press ^C again or type .exit)
>

E:\work\cosmiconfig
$ C:

C:\Users\cdyf
$ D:

D:\Program Files\cmder\config\profile.d
```

davidtheclark: I'm not sure what this means ... Are you saying that `path.resolve()` is not working when you access a file in a drive different from where `process.cwd()` is

lib/createExplorer.js
gucong3000: And this way to avoid the possibility of infinite loop

lib/createExplorer.js
gucong3000: In Windows `path.resolve("E:")` return a string path that same with `process.cwd()`.
And `process.cwd().split(path.sep)` return `[ 'E:', 'work', 'cosmiconfig' ]`

So in last step of loop, `path.resolve("E:")` will run. It make a infinite loops.

There is a demo
```javascript
function walk(dir) {
	dir = path.resolve(dir).split(path.sep);
	console.log(dir);
	walk(dir.slice(0, -1).join(path.sep));
}

walk(process.cwd());
```
It will run indefinitely on Windows.

lib/createExplorer.js
gucong3000: @davidtheclark
It has changed, I think it's much better.

lib/createExplorer.js
davidtheclark: Thanks @gucong3000, I understand now!

davidtheclark:
@gucong3000 Thanks again. In the future please try to limit your PRs to fixing a single precise problem that you've explained. This PR should have been much smaller if all it needed to accomplish was check `splitSearchPath.length > 1` before continuing the search.

PR #36, opened 2016-10-13
Add Node 0.12 support

### LINKED ISSUE #35: Remove const for Safari and node.js 0.12 support
I saw that there is `const` in this module code. PostCSS will use `postcss-load-config`, so it will have `cosmiconf` in dependencies.

And here we have few problems:

* lack node.js 0.12 support. It is not a big deal, because it will be depreacted soon.
* lack of Safari support. Some users use PostCSS in client-side. Of course, cosmiconfig will not be *executed* in browser. But Safari will not even load JS with `const`.

Of course, we could ask client-side users to use Babel. But because problems are only with `const`, I think it is too complicated :). `var` is not so trendy, but it works and it is easy :).

@davidtheclark what do you think about it?

davidtheclark:
Closes #35.

To add Node 0.12 support:
- Replaced `const` and `let` with `var`.
- Replaced arrow functions with regular old anonymous functions.
- Replaced template literals with sad string concatenation.
- Added `object-assign` ponyfill.

@ai: Before I merge and release, would you please try this branch in `postcss-load-config` to double-check that I caught everything you need?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,11 @@
 {
   "extends": [
     "eslint-config-davidtheclark-node"
-  ]
+  ],
+  "rules": {
+    "no-var": "off",
+    "prefer-const": "off",
+    "prefer-arrow-callback": "off",
+    "node/no-unsupported-features": ["error", {"version": 0.12}],
+  }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
 node_js:
+  - "0.12"
   - "4"
   - "6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: node_js
-matrix:
-  include:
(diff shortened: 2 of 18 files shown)


ai:
Wow! You has really fast.

We could test right now on Travis CI. My tests are inside postcss-loader :).

But maybe @michael-ciniawsky could helps us with tests? Could you release new beta with master’s cosmiconfig in `dependencies`?

davidtheclark:
@ai: So this is one of the reasons I dropped Node 0.12 support — now I have a situation where Travis will fail unless I downgrade ESLint. Did you figure out a good work around for this?

davidtheclark:
That tool node-version-check works well for this problem ESLint-Node-version problem.

ai:
`node-version-check` looks awesome!

ai:
I think you could release it as a patch update, because there is no public API changes. Thanks for quick solution.

PR #37, opened 2016-11-13
Don't throw on duplicate keys when parsing YAML

silverwind:
The [`json`](https://github.com/nodeca/js-yaml#safeload-string---options-) option of js-yaml allows duplicate keys to be parsed without throwing, making it compatible with `JSON.parse` in this regard.

Resolves https://github.com/stylelint/stylelint/issues/2051.


davidtheclark:
My understanding is that this is no longer the intended solution to https://github.com/stylelint/stylelint/issues/2051.

Issue #41, opened 2016-11-17
Add support for plugins path lookup


gucong3000:
- http://stylelint.io/user-guide/configuration/#plugins
- http://babeljs.io/docs/plugins/#pluginpreset-paths
- http://eslint.org/docs/user-guide/configuring#configuring-plugins
- http://eslint.org/docs/user-guide/configuring#specifying-parser

davidtheclark:
@gucong3000 I think I'm going to close this issue, #40, #39, and #42.

cosmiconfig has a specific purpose: to find and load configuration files. If you want to do this *and* do other things, like plugins/extends/ignores, I don't see a compelling reason that all of that should be built into the same library. But I can think of a reason or two why *not* to build all that into cosmiconfig, the main one being that libraries are typically better (more stable, more well-tested, more maintainable) the more *focused* they are. 

Why not write a new library that resolves paths for arbitrarily designated properties?

davidtheclark:
After looking at the proof-of-concept in #42, I still think this feature is better suited to its own package than as an extension to cosmiconfig.

PR #42, opened 2016-11-23
Module path lookup

gucong3000:
Add support for plugin modules path lookup #41
- [x] Module path.
  - [x] Relative to configuration file.
- [x] NPM module name support. #45
  - [x] Module that locally-installed in user's project
  - [x] Module that globally-installed
  - [x] **fallback** -- Module not installed by user, locally-installed in tool's directory that dependence `cosmiconfig`.
  - [x] Shorthand of module name


davidtheclark:
@gucong3000 Hold on ... I'm not sure I agree that this feature should be added. Please wait until I have time to think about it — probably late this week, early next.

gucong3000:
I removed some of the demand.

Now only need to find the file path.

gucong3000:
@davidtheclark 
Completed, please review the code.

davidtheclark:
@gucong3000 As expressed before, I don't think I am interested in adding a feature where certain fields (e.g. `extends` or `plugins`) are treated in a special manner. However, I can see how users of cosmiconfig might frequently want a `resolveModules` function that does what you're aiming for. So what do you think of *not* building this in automatically but instead *exposing a new function*, `resolveModules`, that is a utility users could use separately?

I'm thinking of an API like this:

```js
var explorer = cosmiconfig(yourModuleName);
explorer.resolveModule({
  moduleName: /* string */, 
  configPath: /* string */,
}).then((absoluteModulePath) => ..);
```

It would do these things (mostly from your list above):

- Return absolute paths.
- Resolve relative paths relative to `configPath`.
- Resolve module paths:
  - Understand shorthand module names based on cosmiconfig's `yourModuleName` argument.
  - Look for a locally-installed module.
  - Look for a module installed by the dependency that uses cosmiconfig.
  - Look for a globally-installed module.

Are you interested in shifting your work to this API?

gucong3000:
@davidtheclark Sure. i will work on this.

gucong3000:
@davidtheclark
Completed.

-  [x] Modify document
-  [x] Modify API
-  [x] Add test cases

gucong3000:
@davidtheclark
Completed, please review the code.

gucong3000:
I'd like add support for `ignoreFile` & `ignore` in next PR.

davidtheclark:
@gucong3000 Great! I will try to review this within the next week.

gucong3000:
Once upon a time, there was a sleeping beauty, waiting for her prince to kiss her.

davidtheclark:
Haha. Sorry for the delay — I have been extremely busy, and this hasn't been a priority for me. I have not forgotten, though, so I'll try to get to it soon.

davidtheclark:
@gucong3000: Sorry for all the delay. On reconsidering this, I actually think we should not include this feature in cosmiconfig. This could be a package of its own — but I don't see a reason for it to be coupled with cosmiconfig. I suggest making that package, transferring your work here into it!

Again, thanks for your help and sorry that I've been confusing about this.

chrisblossom:
I've created a package that does this [chrisblossom/resolve-with-prefix](https://github.com/chrisblossom/resolve-with-prefix). Although I think it is better to use [chrisblossom/ex-config](https://github.com/chrisblossom/ex-config) to handle the processing and extending of loaded configurations, which uses [resolve-with-prefix](https://github.com/chrisblossom/resolve-with-prefix) under the hood.

PR #44, opened 2016-11-24
fix err ENOENT #44

gucong3000:


davidtheclark:
We do want to throw this error when a file does not exist. Consumers of cosmiconfig (in this case, stylefmt) need to anticipate the error and act accordingly.

Issue #48, opened 2016-12-19
Duplicate keys throw an error with an unhelpful message 


JaKXz:
I am working on stylelint webpack plugin and I think it would be appropriate for stylelint itself to inform users that they have duplicate keys in their `.stylelintrc`. I have been trying to get a PR going but I haven't had success running the tests on master yet. 

Raising the issue here because I don't think that it's difficult to fix but I'm kinda blocked right now.

JaKXz:
@davidtheclark what's blocking me from fixing this is the tests failing from a clean state on master when I clone and run `yarn install` on my system with the latest stable node (v7.4.0 at time of writing). Does this package depend on a specific version of node?

couple questions: I noticed that node v0.12 support was added recently even though the LTS support dropped as of December. Are there plans to move forward again? 
- regardless, I'm quite sure that the tests should be written in ES2015+ since AVA transpiles them anyway.

davidtheclark:
@JaKXz How are the tests failing? What is the error? Have you tried seeing if the tests pass on Travis with Node 7 (by changing the `.travis.yml` file in a PR)?

I just tried `rm -rf node_modules && yarn install && npm test` and had no problem. I am on Node v4.7.0, Yarn v0.18.1.

> Does this package depend on a specific version of node?

[As specified in the `package.json`](https://github.com/davidtheclark/cosmiconfig/blob/master/package.json#L54), this package should work on Node 0.12 and higher.

> I noticed that node v0.12 support was added recently

It was *re-added*, because we removed support but then some users wanted it back, and it was easy to restore.

> Are there plans to move forward again?

I don't understand "move forward" here. There shouldn't be a problem with higher versions of Node. Maybe your error message will clarify things.

> I'm quite sure that the tests should be written in ES2015+ since AVA transpiles them anyway.

I disagree. I'd rather have the codebase stick to one feature set than write the tests differently than the rest of the code.

JaKXz:
@davidtheclark the failures are here: https://travis-ci.org/davidtheclark/cosmiconfig/jobs/189389524

I had spent some time a few weeks ago attempting to debug the failures locally but made no progress without context. 

re: 
> I'd rather have the codebase stick to one feature set than write the tests differently than the rest of the code

[using `t` for all assertions](https://github.com/avajs/ava/issues/1031) is the compromise that I'm suggesting, then.

JaKXz:
[...]

PR #60, opened 2017-04-13
tests: failing test for duplicate keys

JaKXz:
@davidtheclark finally found a few minutes to write a failing test for #48.

Unfortunately to fix this it looks like we have to go a couple packages downstream... what would you suggest?

anyway I'm just opening this to keep up some notion of progress


davidtheclark:
@JaKXz: I don't think there's any need to keep an open PR only for a notion of progress :)

If it's not going anywhere at the moment, I'd prefer to close it, and re-open when there's something actionable.

Issue #59, opened 2017-04-07
Look in user's home even if  not in the path


sidferreira:
Hi! I started the following question on https://github.com/stylelint/stylelint/issues/2479 discussing about ensure that the user's home is checked for configuration files.

The scenario is the following: 
I have a 128gb mounted on my card reader. That card is never removed and that's why I place almost all my projects there. 

As it isn't mounted inside the user's home it will fail to find the configuration on the user's home.

This is a small suggestion to double check the system's path to be sure it is checked. I didn't made a PR because I'm short in time at this moment to look the best place to change the code. If you guys agree it's a good idea, you can assign it to me and I'll look for it.

davidtheclark:
> If you guys agree it's a good idea, you can assign it to me and I'll look for it.

I think it's a good idea!

davidtheclark:
This is possible now with the `searchPlaces` option, so the feature request should be shifted to cosmiconfig consumers (like stylelint) rather than cosmiconfig.

PR #61, opened 2017-04-16
Adds the user home to the search path

sidferreira:
Following to #59, added the functionality that look into user's home for the config files.
That code broke 2 tests, which I fixed and added alternate versions to include the user's home to the tests.

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,6 +17,7 @@ module.exports = function (moduleName, options) {
     rcStrictJson: false,
     stopDir: oshomedir(),
     cache: true,
+    ensureUserHome: true,
   }, options);
 
   if (options.argv && parsedCliArgs[options.argv]) {
--- a/lib/createExplorer.js
+++ b/lib/createExplorer.js
@@ -6,12 +6,14 @@ var loadPackageProp = require('./loadPackageProp');
 var loadRc = require('./loadRc');
 var loadJs = require('./loadJs');
 var loadDefinedFile = require('./loadDefinedFile');
+var oshomedir = require('os-homedir');
 
 module.exports = function (options) {
   // These cache Promises that resolve with results, not the results themselves
   var fileCache = (options.cache) ? new Map() : null;
   var directoryCache = (options.cache) ? new Map() : null;
   var transform = options.transform || identityPromise;
+  var mustEnsureUserHome = false;
 
   function clearFileCache() {
     if (fileCache) fileCache.clear();
@@ -41,6 +43,7 @@ module.exports = function (options) {
     if (!searchPath) return Promise.resolve(null);
 
     var absoluteSearchPath = path.resolve(process.cwd(), searchPath);
+    mustEnsureUserHome = options.ensureUserHome && (process.cwd().indexOf(oshomedir) !== 0);
 
     return isDirectory(absoluteSearchPath)
       .then(function (searchPathIsDirectory) {
@@ -51,7 +54,7 @@ module.exports = function (options) {
       });
   }
 
-  function searchDirectory(directory) {
+  function searchDirectory(directory, lastSearchDirectory) {
     if (directoryCache && directoryCache.has(directory)) {
       return directoryCache.get(directory);
     }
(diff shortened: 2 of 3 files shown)


lib/createExplorer.js
```diff
@@ -51,7 +54,7 @@ module.exports = function (options) {
       });
   }
 
-  function searchDirectory(directory) {
+  function searchDirectory(directory, lastSearchDirectory) {
```
davidtheclark: Do we need this argument? Maybe there is some logic that can avoid it. For example, what if we said:

If there is no next directory, 
  and this is not `stopDir`, 
  and this is not the home directory, 
then check the home directory.

Would that have the same effect without requiring `mustEnsureUserHome` or `lastSearchDirectory`?

lib/createExplorer.js
```diff
@@ -77,7 +80,13 @@ module.exports = function (options) {
           ? splitPath.slice(0, -1).join(path.sep)
           : null;
 
-        if (!nextDirectory || directory === options.stopDir) return null;
+        if (!nextDirectory || directory === options.stopDir || lastSearchDirectory) {
+          if (mustEnsureUserHome && directory !== oshomedir()) {
+            return searchDirectory(oshomedir(), true);
```
davidtheclark: Whatever the logic ends up being, can we avoid calling this `oshomedir()` function multiple times, and instead save the value in a variable?

test/failed-directories.test.js
```diff
@@ -6,11 +6,16 @@ var path = require('path');
...
 var _ = require('lodash');
 var cosmiconfig = require('..');
+var oshomedir = require('os-homedir');
 
 function absolutePath(str) {
   return path.join(__dirname, str);
 }
 
+function atHomeDir(str) {
+  return oshomedir() + '/' + str;
```
davidtheclark: I think this `/` is causing tests to fail on windows.

test/failed-directories.test.js
```diff
@@ -83,7 +89,71 @@ test('do not find file, and give up', function (assert) {
...
+      case atHomeDir('foo.config.js'):
+        callback({ code: 'ENOENT' });
+        break;
+      default:
+        callback(new Error('irrelevant path ' + searchPath));
+    }
+  });
+
+  var loadConfig = cosmiconfig('foo', {
+    stopDir: absolutePath('.'),
```
davidtheclark: If this is the config, I don't understand why we're looking in the home directory at all. The searching should stop in this very `test/` directory, right?

test/failed-directories.test.js
```diff
@@ -129,6 +200,61 @@ test('stop at stopDir, and give up', function (assert) {
...
+      case atHomeDir('foo.config.js'):
+        callback({ code: 'ENOENT' });
+        break;
+      default:
+        callback(new Error('irrelevant path ' + searchPath));
+    }
+  });
+
+  var loadConfig = cosmiconfig('foo', {
+    stopDir: absolutePath('a'),
```
davidtheclark: Same question as above.

index.js
```diff
@@ -17,6 +17,7 @@ module.exports = function (moduleName, options) {
     rcStrictJson: false,
     stopDir: oshomedir(),
     cache: true,
+    ensureUserHome: true,
```
SidFerreira: I added it because of the unit tests. I wanted to have a way to control if the original behavior was working the same way.

lib/createExplorer.js
SidFerreira: I'm wondering about other options. The problem is: if you start search directory one last time, only for the home, the `nextDirectory` variable WILL have a value and will be DIFFERENT than `stopDir`, so, it will keep searching on other folders. The goal is to search on user's home only once and don't go to the next directory. We can change it to ask if should go down to the next directory or not too.

test/failed-directories.test.js
SidFerreira: Sorry, but, which `text/` directory?

test/failed-directories.test.js
SidFerreira: Sorry, but, which `text/` directory?

davidtheclark: `test/`, not `text/` :)

This line says that the `stopDir` is going to be this directory, `test/`. Which means that cosmiconfig should stop at this point, not continue on to the home directory — right?

lib/createExplorer.js
davidtheclark: > the nextDirectory variable WILL have a value and will be DIFFERENT than stopDir, so, it will keep searching on other folders

When we decide to stop searching, though, aren't we comparing `directory` to `stopDir` — not `nextDirectory`? So if the value of `directory` is the home directory, it will *first* search that directory, then check and see if it's the same as `stopDir`, then stop.

I guess the real way to play with the algorithm would be to get strong tests in place first, then make sure they pass.

davidtheclark:
Closing as stale. If somebody would like to revisit this functionality, feel free to open a new PR.

PR #64, opened 2017-04-24
Replace isDirectory with is-directory from npm. Closes #63.

### LINKED ISSUE #63: Should `ENOENT` be suppressed when looking up a non-existent location?
Ref: https://github.com/postcss/postcss-cli/issues/122

cosmiconfig will throw an error when trying to resolve config from a postcss-cli input coming in from `stdin`. This is because the filepath is set to whatever the `process.cwd()` is plus `stdin` (e.g. `/path/to/project/stdin`).

I noticed that the current `isDirectory` function in https://github.com/davidtheclark/cosmiconfig/blob/de81cf6f4bd8f18f4c36255b8a9f13733e3cc7f7/lib/createExplorer.js#L100 doesn't make any exception for `ENOENT` type errors, however the utility `is-directory` from npm [does](https://github.com/jonschlinkert/is-directory/blob/a34146c4dd285680641da47a4fcdfac807dc101a/index.js#L28). Do you see any problems with migrating to `is-directory` instead?

ben-eb:


ben-eb:
Thanks for the quick resolution. 👍

PR #67, opened 2017-05-28
feat: Add option to load config in sync mode

sudo-suhas:
This pull request allows `cosmiconfig` to load config files synchronously. I have updated the tests so that each test runs for both sync and async modes. I have also updated the README as well. I'd be really grateful if you could accept the pull request.

<details>
<summary>My use case</summary>

I am trying to author a library for building elasticsearch queries using a simpler syntax:
```js
// input
'["discount"] is false or (["psngr_cnt"] > 81 and ["booking_mode"] contains "Airport")'

// output
{
  "bool": {
    "should": [
      { "term": { "discount": false } },
      {
        "bool": {
          "must": [
            {
              "range": { "psngr_cnt": { "gt": 81 } }
            },
            {
              "match": { "booking_mode": "Airport" }
            }
          ]
        }
      }
    ]
  }
}
```

I want to allow the user to tweak the queries generated by adding a config file. The library has a synchronous API and hence needs the config to be loaded in sync as well.
</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,11 +179,12 @@ If `true`, config will be loaded synchronously.
 
 ##### transform
 
-Type: `Function` returning a Promise(`sync` is false)
+Type: `Function`
 
-A function that transforms the parsed configuration. Receives the result object with `config` and `filepath` properties, and must return a Promise that resolves with the transformed result.
+A function that transforms the parsed configuration. Receives the result object with `config` and `filepath` properties.
 
-If the `sync` option is passed as `true`, transform need not be a function which returns a promise.
+If the option `sync` is `false`(default), the function must return a Promise that resolves with the transformed result.
+Otherwise, `transform` should be a synchronous function which returns the transformed result.
 
 The reason you might use this option instead of simply applying your transform function some other way is that *the transformed result will be cached*. If your transformation involves additional filesystem I/O or other potentially slow processing, you can use this option to avoid repeating those steps every time a given configuration is loaded.
 
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ For example, if your module's name is "soursocks," cosmiconfig will search out c
 - a `soursocks.config.js` file exporting a JS object (anywhere down the file tree)
 - a CLI `--config` argument
 
-cosmiconfig continues to search in these places all the way down the file tree until it finds acceptable configuration (or hits the home directory). And it does all this asynchronously, so it shouldn't get in your way.
+cosmiconfig continues to search in these places all the way down the file tree until it finds acceptable configuration (or hits the home directory). And it does all this asynchronously by default, so it shouldn't get in your way.
 
 Additionally, all of these search locations are configurable: you can customize filenames or turn off any location.
 
@@ -48,6 +48,8 @@ explorer.load(yourSearchPath)
 
 The function `cosmiconfig()` searches for a configuration object and returns a Promise,
 which resolves with an object containing the information you're looking for.
+
+You can also pass option `sync: true` to load the config synchronously.
 
 So let's say `var yourModuleName = 'goldengrahams'` — here's how cosmiconfig will work:
 
@@ -168,11 +170,20 @@ Default: `true`
 
 If `false`, no caches will be used.
 
+##### sync
+
+Type: `boolean`
+Default: `false`
+
+If `true`, config will be loaded synchronously.
+
 ##### transform
 
-Type: `Function` returning a Promise
+Type: `Function` returning a Promise(`sync` is false)
 
 A function that transforms the parsed configuration. Receives the result object with `config` and `filepath` properties, and must return a Promise that resolves with the transformed result.
+
+If the `sync` option is passed as `true`, transform need not be a function which returns a promise.
 
 The reason you might use this option instead of simply applying your transform function some other way is that *the transformed result will be cached*. If your transformation involves additional filesystem I/O or other potentially slow processing, you can use this option to avoid repeating those steps every time a given configuration is loaded.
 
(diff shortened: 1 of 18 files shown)


sudo-suhas:
pretty-please

davidtheclark:
@sudo-suhas: This is a very significant change — sorry, I don't have a ton of time for OSS right now. I'll try to get to it at some point, though!

sudo-suhas:
@davidtheclark I understand. I'll just use it with github URL to my fork until you have the time.

sudo-suhas:
@davidtheclark Any chance you could take a look at this now? I'd really appreciate it.

README.md
```diff
@@ -168,12 +170,21 @@ Default: `true`
...
+If `true`, config will be loaded synchronously.
+
 ##### transform
 
-Type: `Function` returning a Promise
+Type: `Function` returning a Promise(`sync` is false)
 
 A function that transforms the parsed configuration. Receives the result object with `config` and `filepath` properties, and must return a Promise that resolves with the transformed result.
 
+If the `sync` option is passed as `true`, transform need not be a function which returns a promise.
```
davidtheclark: More specifically, it `transform` needs to be a `synchronous` function, right?

lib/createExplorer.js
```diff
@@ -32,56 +32,88 @@ module.exports = function (options) {
...
+          if (result || !options.js) return result;
+          return loadJs(path.join(directory, options.js), options);
+        })
+        .then(function (result) {
+          if (result) return result;
+
+          var splitPath = directory.split(path.sep);
+          var nextDirectory = (splitPath.length > 1)
+            ? splitPath.slice(0, -1).join(path.sep)
+            : null;
```
davidtheclark: I want to avoid duplicating lines like this here and on lines 106+. Can we make a function for it?

lib/loadJs.js
```diff
@@ -3,13 +3,17 @@
...
     return {
       config: requireFromString(content, filepath),
       filepath: filepath,
     };
-  });
+  }
+
+  return (!options.sync)
+    ? readFile(filepath).then(parseJsFile)
+    : parseJsFile(readFile.sync(filepath));
```
davidtheclark: I like this method of creating the `parse...` functions and only differentiating sync and async in these few lines!

lib/loadRc.js
```diff
@@ -73,13 +87,31 @@ module.exports = function (filepath, options) {
...
+        .then(parseJsRcFile)
+      : parseJsRcFile(
+          parseYamlRcFile(
+            parseYmlRcFile(
+              parseJsonRcFile(
+                readRcFile('json')
+              )
+            )
+          )
+        );
```
davidtheclark: For this and similarly large ternaries — can we favor early returns, instead?

sudo-suhas:
Awesome! Thank you so much 😃. I'll try to do the requested changes by tomorrow.

sudo-suhas:
Hey @davidtheclark, I was taking a fresh look at the code and I can think of 2 possible refactors. Let me know what you think.
In quite a few places, there are exit clauses for functions:
 - [`createExplorer.js`#L61](https://github.com/davidtheclark/cosmiconfig/blob/master/lib/createExplorer.js#L61)
 - [`createExplorer.js`#L65](https://github.com/davidtheclark/cosmiconfig/blob/master/lib/createExplorer.js#L65)
 - [`createExplorer.js`#L69](https://github.com/davidtheclark/cosmiconfig/blob/master/lib/createExplorer.js#L69)
- [`loadRc.js`#L10](https://github.com/davidtheclark/cosmiconfig/blob/master/lib/loadRc.js#L10)
- [`loadRc.js`#L47](https://github.com/davidtheclark/cosmiconfig/blob/master/lib/loadRc.js#L47)

Would you consider using [`p-if`](https://github.com/sindresorhus/p-if) for this? But it has the downside of needing node version 4 or above.

In both [`createExplorer.js`](https://github.com/sudo-suhas/cosmiconfig/blob/feat_load_sync/lib/createExplorer.js#L68-L115) and [`loadRc.js`](https://github.com/sudo-suhas/cosmiconfig/blob/feat_load_sync/lib/loadRc.js#L92-L106), we have a series of functions that need to be chained. I can think of a way to abstract this into a function runner. 

Pseudo code:
```js
let res = sync ? null : Promise.resolve();

funcs.forEach(func => {
  if (isPromise(res)) res = res.then(func);
  else res = func(res);
})

```

README.md
sudo-suhas: You are right. While it wouldn't break anything, having a promise returning function for `transform` would render the `sync` option moot.

sudo-suhas:
I didn't notice this earlier - [`createExplorer.js`#L11](https://github.com/davidtheclark/cosmiconfig/blob/master/lib/createExplorer.js#L11).

> `// These cache Promises that resolve with results, not the results themselves`

While this won't be true for the sync execution, just wanna confirm with you that this cannot cause any issues. As far as I understand, the cache will be tied to the options and instantiating a new loader should get a new cache as well.

davidtheclark:
> As far as I understand, the cache will be tied to the options and instantiating a new loader should get a new cache as well.

Yeah, that should be the case. 👍

sudo-suhas:
@davidtheclark The changes should address the concerns you had. If you are okay with `funcRunner`, I'll add test cases for it as well.

davidtheclark:
@sudo-suhas: Looks good! Yep, tests for `funcRunner` would finish it up.

sudo-suhas:
@davidtheclark Thanks for merging this in. Could I make another PR with updated dependencies before you release it?

davidtheclark:
Ok

davidtheclark:
Released in 2.2.0.

PR #68, opened 2017-07-09
Fix --config and configPath option

azz:
`options.configPath` was never read, so the `--config` flag didn't work.


davidtheclark:
@azz: Can you please add a test that shows how it was broken before, and how this change fixes it?

azz:
Can do! Test fails when the change is commented out.

PR #74, opened 2017-07-22
Fix transform not being run

### LINKED ISSUE #72: 2.2.0 possible breaks stylelint
Sorry, @davidtheclark, for issue like that.

I haven't investigated, but it looks like `2.2.0` breaking stylelint:

https://github.com/stylelint/stylelint/issues/2768
https://github.com/stylelint/stylelint/issues/2769

sudo-suhas:
### Changelog

- Revert the [revert](https://github.com/davidtheclark/cosmiconfig/commit/31467d4005fc9771a67af9614adc68ff76c7137b). Let me know if there's a better way.
- Fix issue with transform function resolution in `createExplorer.js`, details below
- Add test for transform option

### Details

The transform function was not being resolved correctly - [`lib/createExplorer.js#L16`](https://github.com/davidtheclark/cosmiconfig/blob/b57291583fcda08edad2aff43257de7fcb3c3938/lib/createExplorer.js#L16)
Here's a simple demo for what was going wrong:
```
λ node
> let sync = false

> 'someValidFunc' || (sync) ? 'identitySync' : 'identityPromise'
'identitySync'

> 'someValidFunc' || (sync ? 'identitySync' : 'identityPromise')
'someValidFunc'
```

And because the loaded config was not being transformed properly, stylelint was breaking(stylelint/stylelint#2768).

Closes #72

### DIFF
--- a/lib/createExplorer.js
+++ b/lib/createExplorer.js
@@ -10,10 +10,11 @@ var funcRunner = require('./funcRunner');
 
 module.exports = function (options) {
   // When `options.sync` is `false` (default),
-// these cache Promises that resolve with results, not the results themselves.
+  // these cache Promises that resolve with results, not the results themselves.
   var fileCache = (options.cache) ? new Map() : null;
   var directoryCache = (options.cache) ? new Map() : null;
-  var transform = options.transform || (options.sync) ? identitySync : identityPromise;
+  var defaultTransform = options.sync ? identitySync : identityPromise;
+  var transform = options.transform || defaultTransform;
 
   function clearFileCache() {
     if (fileCache) fileCache.clear();
--- a/lib/createExplorer.js
+++ b/lib/createExplorer.js
@@ -6,12 +6,14 @@ var loadPackageProp = require('./loadPackageProp');
 var loadRc = require('./loadRc');
 var loadJs = require('./loadJs');
 var loadDefinedFile = require('./loadDefinedFile');
+var funcRunner = require('./funcRunner');
 
 module.exports = function (options) {
-  // These cache Promises that resolve with results, not the results themselves
+  // When `options.sync` is `false` (default),
+// these cache Promises that resolve with results, not the results themselves.
   var fileCache = (options.cache) ? new Map() : null;
   var directoryCache = (options.cache) ? new Map() : null;
-  var transform = options.transform || identityPromise;
+  var transform = options.transform || (options.sync) ? identitySync : identityPromise;
 
   function clearFileCache() {
     if (fileCache) fileCache.clear();
@@ -36,23 +38,31 @@ module.exports = function (options) {
       if (fileCache && fileCache.has(absoluteConfigPath)) {
         return fileCache.get(absoluteConfigPath);
       }
-      var result = loadDefinedFile(absoluteConfigPath, options)
-        .then(transform);
+      var result = (!options.sync)
+        ? loadDefinedFile(absoluteConfigPath, options)
+          .then(transform)
+        :  transform(loadDefinedFile(absoluteConfigPath, options));
       if (fileCache) fileCache.set(absoluteConfigPath, result);
       return result;
     }
 
-    if (!searchPath) return Promise.resolve(null);
+    if (!searchPath) return (!options.sync) ? Promise.resolve(null) : null;
 
     var absoluteSearchPath = path.resolve(process.cwd(), searchPath);
 
-    return isDirectory(absoluteSearchPath)
-      .then(function (searchPathIsDirectory) {
-        var directory = (searchPathIsDirectory)
+    return (!options.sync)
+      ? isDirectory(absoluteSearchPath)
+        .then(function (searchPathIsDirectory) {
+          var directory = (searchPathIsDirectory)
+            ? absoluteSearchPath
+            : path.dirname(absoluteSearchPath);
+          return searchDirectory(directory);
+        })
+      : searchDirectory(
+        isDir.sync(absoluteSearchPath)
           ? absoluteSearchPath
-          : path.dirname(absoluteSearchPath);
-        return searchDirectory(directory);
-      });
+          : path.dirname(absoluteSearchPath)
+      );
   }
 
   function searchDirectory(directory) {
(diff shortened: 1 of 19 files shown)


sudo-suhas:
@davidtheclark travis build did not run for this.. And in travis build requests, I can see 'Build type disabled via repository settings (pull_request).' Should we enable builds for PRs again?

davidtheclark:
Yes let's switch that back on in Travis. I did a t mistakingly.

sudo-suhas:
> Yes let's switch that back on in Travis
Done.

lib/createExplorer.js
```diff
@@ -10,10 +10,10 @@ var funcRunner = require('./funcRunner');
 
 module.exports = function (options) {
   // When `options.sync` is `false` (default),
-// these cache Promises that resolve with results, not the results themselves.
+  // these cache Promises that resolve with results, not the results themselves.
   var fileCache = (options.cache) ? new Map() : null;
   var directoryCache = (options.cache) ? new Map() : null;
-  var transform = options.transform || (options.sync) ? identitySync : identityPromise;
+  var transform = options.transform || (options.sync ? identitySync : identityPromise);
```
davidtheclark: I'd suggest we clarify this syntax more to avoid misunderstandings in the future. I certainly missed this subtlety. How about:

```js
var defaultTransform = options.sync ? identitySync : identityPromise;
var transform = options.transform || defaultTransform;
```

test/successful-files.test.js
```diff
@@ -24,7 +24,7 @@ function makeFileTest(file) {
     try {
       var result = loadConfigSync(null, filePath);
       doAsserts(assert, result, filePath);
-      loadConfig(null, absolutePath(file)).then(function (result) {
+      loadConfig(null, filePath).then(function (result) {
```
davidtheclark: What's the reason for this change?

test/successful-files.test.js
```diff
@@ -43,3 +43,45 @@ test('defined YAML config path', makeFileTest('fixtures/foo.yaml'));
...
+      );
+
+      assert.end();
+    }).catch(function (err) {
+      assert.end(err);
+    });
+  } catch (err) {
+    assert.end(err);
+  }
+});
```
davidtheclark: I'd rather have two separate tests, one for sync, one for async —— this one test is kind of difficult to understand.

davidtheclark:
One more thought: Can we add a test to ensure that if your `transform` function errors, those errors make it through, aren't swallowed up?

lib/createExplorer.js
sudo-suhas: I was surprised too. I completely agree with the suggested change. On a side note, there is actually no need to have a different identity func for sync and async. We can use the sync version for both.

test/successful-files.test.js
sudo-suhas: Actually `filePath` is the same as `absolutePath(file)` and is evaluated at the beginning of the higher order function([`test/successful-files.test.js#L19`](https://github.com/sudo-suhas/cosmiconfig/blob/f393c95ac789cc4ceb9c284d3d16fc8f5b46772e/test/successful-files.test.js#L19)).

test/successful-files.test.js
sudo-suhas: I had modified all the tests while I was submitting the feature for sync to test both sync and async. The reason for this was simply code reuse. I will split this one up. But should we be doing that for all tests as well?

sudo-suhas:
> Can we add a test to ensure that if your `transform` function errors, those errors make it through, aren't swallowed up?

Okay. I'll do that as well.

sudo-suhas:
After doing the changes for node 4, I'd like to improve tests as well. I am considering using jest and adding integration tests with stylelint plus code coverage.

test/successful-files.test.js
davidtheclark: I was thinking about doing that for this one because it's new, so easy to make clear. The others from your sync PR are fine for now, I think.

davidtheclark:
> I am considering using jest and adding integration tests with stylelint plus code coverage.

Yeah, I've been using Jest a lot lately and really liking it. And integration tests would be awesome.

davidtheclark:
@sudo-suhas: If you are planning on working on these refactors soon, I'm thinking we should hold off on releasing this `sync` feature for now and then do a major release that includes all the changes. Dependents like stylelint then would not be caught off guard, and if the refactoring introduces any other problems they will not be automatically introduced to end users.

sudo-suhas:
@davidtheclark I am in two minds about this. What you are suggesting does make sense. However, #68 was a rather important fix no? I _was_ planning to work on the refactors over the next week or so. You can take a call on this.

davidtheclark:
@sudo-suhas: We could release just that fix. I can do this today.

sudo-suhas:
The first commit of this PR, a revert commit, includes the changes for that fix. It should be enough to rebase right?

davidtheclark:
Just released 2.2.2 with the `--config` fix. If you rebase your branch and fix any commits, it should be fine, I think.

sudo-suhas:
Surprisingly, conflict only in the changelog.
A small clarification, in the changelog, for `v2.2.1`, it is denoted as `2.2.0-1`. Is this intentional?

davidtheclark:
I adjusted the notation to `2.2.0, 2.2.1`.

> Also, shouldn't the changes available in 2.2.0 be listed?

I don't know ... I feel like that was just a flawed release that should not be used. That's why I grouped those two releases into one entry.

sudo-suhas:
Yea, Okay 👍

sudo-suhas:
@davidtheclark I had noticed that you usually squash and merge PRs. If we were to do for this one, it will be difficult to track the actual changes(other than revert). Should I merge the rest of the commits locally?

davidtheclark:
You can use rebase & merge if you like your commit history! I tend to use squash & merge if the commit history on a branch got ugly or was filled with experiments and false directions. I like rebase & merge if the commit history is clear.

sudo-suhas:
@davidtheclark I am regretting having merged this in. Should we create a `3.0` branch and have all refactoring there instead? I could cherry pick the changes in this PR to that and you could force push master to 69b4cb6b.

sudo-suhas:
@davidtheclark One last thing. I think until 3.0 branch is ready and merged in, the readme should accurately reflect the state of the last release. So shall I add a PR reverting this one or can you force write HEAD to  69b4cb6. While I know that force write is generally frowned upon, in this case, I just want to avoid the noise of revert*3.

davidtheclark:
> the readme should accurately reflect the state of the last release

I don't think we have to worry that much about it. This situation is pretty standard. We could add a note to the top of the readme saying it reflects master here's a link to the README for a previous version. I'd rather do that than Git magic.

sudo-suhas:
Alright 👍 I'll defer to your better judgement and experience here

PR #80, opened 2017-09-06
Add integration tests

sudo-suhas:
This PR aims to add integration tests with stylelint and prettier.

### stylelint

Test scenarios:
_TODO_

### prettier

Test scenarios:
- CLI
  - Resolve config for files in multiple directories using the closest one from their directory.
  - Resolve config using the `--config` option.
  - Resolve config for given file using `--find-config-path` option.
- `prettier.resolveConfig`
  - Resolve config for given folder
  - Resolve config for given file


sudo-suhas:
@davidtheclark I need your help for writing integration tests with `stylelint`. The current test is a POC which does a very minimal test to see that `stylelint` works ok with `cosmiconfig`. I would like to add the more advanced scenarios which would test `cosmiconfig` better.

sudo-suhas:
A small win, when I intentionally introduced the bug which was fixed in #74(transform not being run), this integration test was able to catch it 🎉

sudo-suhas:
I cannot write integration tests for prettier 😞. They use a build step which bundles all dependencies into a single file which is what I get on npm install. Although the mock for cosmiconfig works, prettier does not load it using require. @azz Is it possible to change this? Also, why is this done for prettier? The file is nearly 21k lines!

azz:
Reasoning for that is here: https://github.com/prettier/prettier/issues/1640

If you install `prettier/prettier` from github you can use the non-bundled version.

sudo-suhas:
Thanks @azz! Is it a good idea to use `prettier` directly from github? Aren't there chances that master branch will have bugs?

azz:
I wouldn't say master is any more or less stable than the released version. Often we fix a lot of stuff in master that hasn't made it to npm yet, and our test suite is quite extensive.

But you can install a specific version with e.g. `npm install prettier/prettier#1.6.1`

sudo-suhas:
This helps 😄 Thank you so much @azz!

davidtheclark:
> I would like to add the more advanced scenarios which would test cosmiconfig better.

@sudo-suhas: Can you please say more about the scenarios you'd like to test? I think what you have is a good place to start!

sudo-suhas:
Tests are failing due to `jest-in-case`, should be fixed once Thinkmill/jest-in-case#1 is merged in.

sudo-suhas:
Can you please say more about the scenarios you'd like to test?

@davidtheclark I want to have 4 sets of configs and CSS files which I can tests similar to prettier. Specifically, that the config is loaded from `.stylelintrc`/`package.json`/`stylelint.config.js` and the file adjacent to it(or in a parent folder) is linted with it. I'll try to take a look at tests in `stylelint` later today. Maybe that will give me what I am looking for.

sudo-suhas:
Builds on appveyor fail because of processor version mismatch in the snapshot generated by stylelint. Switching this to use yarn for deterministic installs on appveyor.

sudo-suhas:
😳 Even that didn't work. Now I need help.

davidtheclark:
@sudo-suhas: My impression is that these tests are turning out to be more trouble than they're worth. And my experience with integration tests is that such troubles might compound over time. What do you think about abandoning integration tests and instead reviewing code coverage?

The stylelint failure that sparked this quest would have been prevented if there was unit test coverage on one additional branch of code.

azz:
We have tests in prettier for a lot of the use cases you mentioned here:

https://github.com/prettier/prettier/blob/master/tests_integration/__tests__/config-resolution.js

If you'd like to add extra tests to prettier to cover other edge cases that'd be welcome.

sudo-suhas:
@azz I actually used the file you linked to write the tests. The intent was being able to test the unreleased version of cosmiconfig with prettier. Meaning, we should detect the bug _before_ we release a version and not _after_. The tests in prettier are definitely useful and necessary.

sudo-suhas:
@davidtheclark I have added the unit tests and raised the code coverage to 100% in #83. It seems like a surprising number of people are waiting for the `sync` feature. Let's release `v3.0.0`. 

As for integration tests, did you get a chance to take a look at the tests I added for prettier? I agree that if these problems are going to compound over time, it's not worth it. So please take a look and let me know what you think.

davidtheclark:
@sudo-suhas: If you want to keep the Prettier tests, we can keep them and see how it goes. My hesitation is that they rely on things that Prettier can and should change at will, like `prettier/tests_integration/runPrettier`. We don't want to get into the situation where test failures are not relevant to the code in cosmiconfig, if we don't have to — and I think with good test coverage, we don't have to. Your call!

sudo-suhas:
I agree with you mostly. However, we are using the prettier with a git tag so we would only have to deal with problems if and when we attempt to change the version being used. But yea, not ideal.

So here's what I suggest. We close this PR but I'll keep the branch in my fork. If we ever detect a bug _after_ releasing a version, we come back to this. Sound good?

davidtheclark:
Sounds good!

I would like to take a quick pass through the code today before releasing v3, see if there's any refactoring that comes to mind. I should have time later today. After that we can release v3 — later today or tomorrow. 👍

sudo-suhas:
Closing this until we are sure integration tests are needed.

PR #81, opened 2017-09-06
feat: load package prop when --config is package.json

### LINKED ISSUE #79: --config cannot point to a package.json
Currently, if a config file is passed explicitly, and that config file is a `package.json`, the contents of the `package.json` file are loaded instead of a single property within it.

https://github.com/davidtheclark/cosmiconfig/blob/334065d9f7b557170a38b3f920bf5e90cecfab72/lib/createExplorer.js#L30-L39

Would it make sense to have a special case for `package.json` that calls `loadPackageProp` instead?

Context: https://github.com/prettier/prettier/issues/2704

azz:
Closes #79


test/caches.test.js
```diff
@@ -32,7 +32,7 @@ beforeAll(() => {
       case absolutePath('a/b/c/foo.config.js'):
         throw { code: 'ENOENT' };
       case absolutePath('a/b/package.json'):
-        return '{ "foundInB": true }';
+        return '{ "foo": { "foundInB": true } }';
```
davidtheclark: Why did this test need to change?

azz: Because this test checks the contents of the configuration, which has changed.

https://github.com/azz/cosmiconfig/blob/ce99ddee32b71658a5d08b02778fe2430ce8bdb3/test/caches.test.js#L140-L150

azz:
Done.

davidtheclark:
Thanks @azz, looks great.

@sudo-suhas: Do you want to merge this now?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.2.3
```
sudo-suhas: The version number isn't correct. This will be release in `3.0.0`. Merging it and updating this to `master` to denote all changes currently in master.

calebeby:
When can this feature be released? Thanks!

sudo-suhas:
@calebeby cosmiconfig 3.0, which includes this bug fix, has now been released to npm.

calebeby:
Awesome, thank you so much!

PR #85, opened 2017-09-11
refactor: rm deprecated package os-homedir

sudo-suhas:
The `os-homedir` package [is deprecated](https://github.com/sindresorhus/os-homedir) as the behavior is identical to `os.homedir()` on node 4 and above:
> This is not needed anymore. Use require('os').homedir() instead.

This PR removes the package and uses the `os` package instead.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,10 +1,12 @@
 'use strict';
 require('please-upgrade-node')(require('./package.json'));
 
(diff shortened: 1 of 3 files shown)


davidtheclark:
🍰  Good catch.

PR #86, opened 2017-09-13
Typecheck configPath cli argument. Defer path.resolve call to config lookup runtime.

amiller-gh:
See https://github.com/prettier/prettier/issues/2797 and https://github.com/prettier/prettier/pull/2798 for why this is needed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,6 @@
 require('please-upgrade-node')(require('./package.json'));
 
 const os = require('os');
-const path = require('path');
 const minimist = require('minimist');
 const createExplorer = require('./lib/createExplorer');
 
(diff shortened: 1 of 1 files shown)


davidtheclark:
Thanks @amiller-gh for investigating.

I feel like it's probably a mistake for cosmiconfig to pay attention to a `--config` argument in `process.argv`. It's just too much of an assumption. This fix solves your very particular problem where the value of `--config` is an array — but it's a pretty specific fix, and I suspect that the more general problem is "cosmiconfig doesn't know enough about your program or its context to cherry pick `process.argv` items."

I'm in favor of removing the `--config` feature altogether. Since we're about to cut a major release (v3), we have a good opportunity. 

@sudo-suhas what do you think?

sudo-suhas:
@davidtheclark I agree. Since cosmiconfig is not meant to be directly invoked from the CLI, makes sense to remove it. The caller can include it in the options if they wish. Would you like me to make a PR for this?

davidtheclark:
@sudo-suhas 👍 go for it.

sudo-suhas:
Okay, will do after #87 is merged in to avoid conflicts.

PR #87, opened 2017-09-13
Add Flow typing

### LINKED ISSUE #84: Flow types
@sudo-suhas How do you feel about adding Flow types? 

I was thinking about adding JSDoc-style comments to more of the functions, but that got me thinking that adding Flow types would both document functions _and_ help with type safety. Type checking might be especially helpful now that we have a bunch of functions that sometimes return Promises, sometimes don't.

If you aren't opposed, I could do that at some point today so we can include it in the 3.0 release.

davidtheclark:
Adds Flow types!

Also:
- Adds `babel-eslint` and adds `eslint-plugin-flowtype` for Flow-ESLint compatibility.
- Adds `transform-flow-strip-types` for Flow-Jest compatibility.
- Renames `lib/` to `src/`.
- Adds a `build` script (which runs on `prepublishOnly`) that builds the `dist/` directory.
- Points `package.json`'s `main` field to `dist/index.js`, so consumer doesn't get Flow types.
- Adds `dist/` to the npm files, but removes `src/`.
- Various small changes throughout the codebase to get Flow passing. More significant revisions in `loadRc.js`, where I think the difficulties I was having adding type annotations reflected an overly complex abstraction, so I favored repetition in that spot over the more complex approach.

@sudo-suhas What do you think?

Closes #84


package.json
```diff
@@ -2,19 +2,21 @@
...
   "description": "Find and load configuration from a package.json property, rc file, or CommonJS module",
-  "main": "index.js",
+  "main": "dist/index.js",
   "files": [
     "index.js",
-    "lib"
+    "dist"
   ],
   "scripts": {
     "precommit": "lint-staged && jest",
```
sudo-suhas: We should add the flow types check in pre-commit hook as well.

package.json
```diff
@@ -51,11 +53,17 @@
   "jest": {
```
sudo-suhas: Do we need to [configure parser](https://github.com/prettier/prettier#parser) for prettier?

package.json
```diff
@@ -51,11 +53,17 @@
...
       "index.js"
     ]
   },
+  "babel": {
+    "plugins": [
+      "transform-flow-strip-types"
+    ]
+  },
   "dependencies": {
+    "flow-bin": "^0.54.1",
```
sudo-suhas: Should this be a dev-dependency?

src/createExplorer.js
```diff
@@ -39,43 +57,66 @@ module.exports = function createExplorer(options) {
         return fileCache.get(absoluteConfigPath);
       }
 
-      const load =
-        path.basename(absoluteConfigPath) === 'package.json'
-          ? () => loadPackageProp(path.dirname(absoluteConfigPath), options)
-          : () => loadDefinedFile(absoluteConfigPath, options);
+      let load;
```
sudo-suhas: If you don't mind, could you explain this change?

src/createExplorer.js
```diff
@@ -39,43 +57,66 @@ module.exports = function createExplorer(options) {
...
     }
 
     if (!searchPath) return !options.sync ? Promise.resolve(null) : null;
 
     const absoluteSearchPath = path.resolve(process.cwd(), searchPath);
-    const searchPathDir = resolveDir(absoluteSearchPath, options.sync);
+    const searchPathDir = getDirectory(absoluteSearchPath, options.sync);
 
-    return !options.sync
+    return searchPathDir instanceof Promise
```
sudo-suhas: Is this something flow forces us to do? Could we use `is-promise` for this?

src/getDirectory.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+'use strict';
+
+const path = require('path');
+const isDirectory = require('is-directory');
+
+module.exports = function getDirectory(
+  filename: string,
```
sudo-suhas: nit: Rename this to `filepath`?

src/index.js
```diff
@@ -1,19 +1,38 @@
...
+    transform?: (?Object) => ?Object,
+    configPath?: string,
+  }
+) {
   // Keeping argv parsing here allows to mock `minimist` for different tests.
   // This should not have too much of a negative impact.
   const parsedCliArgs = minimist(process.argv);
 
   options = Object.assign(
+    {},
```
sudo-suhas: nit: This shouldn't be required right?

src/loadRc.js
```diff
@@ -0,0 +1,112 @@
...
+      ? parseJson(content, filepath)
+      : yaml.safeLoad(content, { filename: filepath });
+    return {
+      config: pasedConfig,
+      filepath,
+    };
+  }
+
+  function loadRcWithExtensions() {
+    let foundConfig = null;
```
sudo-suhas: Is there anyway we can avoid this? I am uncomfortable with the idea of anonymous functions modifying this variable. If adding flow is complicating it, can we skip it for this file?

src/index.js
davidtheclark: This is required. `Object.assign` is pretty hard for Flow to understand.

src/loadRc.js
davidtheclark: I prefer the clarity now, with or without Flow. 

> I am uncomfortable with the idea of anonymous functions modifying this variable.

Why is that? 

It seemed to me the alternative is to make each function a little more ambiguous: might receive a string or a Result object or a falsy value. Having the condition that duck types the argument to check if it is a Result object or a string didn't feel any cleaner to me.

src/createExplorer.js
davidtheclark: Oh, we can use `isPromise`, yes, didn't think of that. 

For all of these functions that return `Promise<string> | string`, Flow doesn't have a good way of knowing if it should check of a string or a Promise. So if you just attach `.then()` to the result, it will complain because that could a string, not a Promise. This is essentially the same check (will be true in all the same situations), but gives Flow an opportunity to check what's going on.

package.json
davidtheclark: Yes.

package.json
davidtheclark: The default in Prettier works fine with Flow.

package.json
davidtheclark: Sure.

src/createExplorer.js
davidtheclark: Any particular part of the change?

The change from a ternary to a full if-else block was because I'd prefer not to use ternaries when the different values can be multiple lines. Creating new objects instead of passing the `options` object directly pleases Flow because it knows that nobody is going to mutate the values somewhere else — and doing that made the functions take up a few lines.

src/createExplorer.js
davidtheclark: Tried this out. `isPromise` doesn't provide the same type safety, and I think we don't need it, actually, because we are using real Promises. I can replace other usage of `isPromise` with an `instanceof` check.

davidtheclark:
Thanks for taking a close look, @sudo-suhas. Addressed the feedback so far.

src/createExplorer.js
sudo-suhas: Okay. But IMHO we should avoid `instanceof` checks. Not sure if this could affect us, but with various implementations of promise, an instance check might not be reliable. See https://github.com/petkaantonov/bluebird/issues/628. And the recommended method is to check the presense of `.then` - https://stackoverflow.com/questions/27746304/how-do-i-tell-if-an-object-is-a-promise.

src/loadRc.js
sudo-suhas: Okay. I'll try to refactor this sometime later using [`p-if`](https://github.com/sindresorhus/p-if)

sudo-suhas:
@davidtheclark The tests are failing on appveyor with `Platform not supported` due to `flow-bin`.

src/loadRc.js
davidtheclark: It sounds to me like that might be unnecessary work that will add new complexity and abstractions that make the code harder to read. But if you want to give it a shot, happy to look.

src/createExplorer.js
sudo-suhas: The code earlier was not using `packageProp`. So now if `absoluteConfigPath` is for a `package.json` and `packageProp` is `false`, we return the entire file as the result?

src/loadRc.js
sudo-suhas: I'll try to refactor in a way which is easy to understand. If I am not able to do that, will leave as is.

src/createExplorer.js
davidtheclark: If you'd like to work on getting `isPromise` into these situations, go for it. I doubt we have to worry about this since we are using real Promises — and tests should fail if we aren't. But I totally understand your point.

sudo-suhas:
@davidtheclark Shall we use yarn on appveyor? Either that or we have to install a newer verison of npm to fix the error. It is [available by default](https://yarnpkg.com/lang/en/docs/install-ci/#appveyor-tab) on appveyor. It will be more consistent as well.

davidtheclark:
@sudo-suhas: Yep, I'll add that line. I got Flow to install (https://github.com/flowtype/flow-bin/issues/50 helped), but running into another problem. Working on it ...

davidtheclark:
@sudo-suhas: Yep, I'll add that line. I got Flow to install (https://github.com/flowtype/flow-bin/issues/50 helped), but running into another problem. Working on it ...

src/createExplorer.js
davidtheclark: Yeah — I think this is one case where Flow typing revealed a bug, right? The code earlier was *implicitly* using `packageProp` by passing `options`. Before, we would have tried to access a property in `package.json` that didn't exist, which would error.

Another possible move here would be to throw an error if `configPath` is `package.json` and `packageProp` is not defined.

davidtheclark: Added a clear error message and tests to ensure it comes through.

test/failed-files.test.js
```diff
@@ -151,5 +152,29 @@ describe('cosmiconfig', () => {
...
+      expect(() =>
+        cosmiconfig('foo', { sync: true, packageProp: false }).load(
+          null,
+          path.join(__dirname, 'fixtures/package.json')
+        )
+      ).toThrow(/Please specify the packageProp option/);
+    });
+
+    it('in async mode, rejects if configPath is package.json and packageProp is false', () => {
+      return cosmiconfig('foo', { packageProp: false })
```
sudo-suhas: If you add a `expect.assertions(1)`, there is no need to define the `.then` clause.

sudo-suhas:
@davidtheclark Maybe you saw this already - https://blog.acolyer.org/2017/09/19/to-type-or-not-to-type-quantifying-detectable-bugs-in-javascript/(javascript newsletter).

davidtheclark:
I had not seen that; thanks for sharing it!

PR #88, opened 2017-09-13
ci: Whitelist build branches to avoid duplicate builds

sudo-suhas:
For pull requests which are made from a branch in this repo instead of a fork, travis and appveyor run duplicate builds once for the PR and once for the branch push. By whitelisting branches, we can limit builds to master and PR branches. To be clear, if a branch, with which a PR was made, is updated, that will still trigger a build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,6 @@ node_js:
   - "6"
   - "7"
   - "8"
+branches:
+  only:
+    - master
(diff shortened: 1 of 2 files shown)


davidtheclark:
👍

chrisblossom:
I think this should be reverted. If you are working off of a fork you won't cause any duplicate builds. So only @davidtheclark should be causing duplicate builds.

Maybe there is a better solution that I am unaware of though?

PR #92, opened 2017-09-17
Bump dependency versions

### LINKED ISSUE #90: JS config never gets garbage collected.
I discovered this issue while trying to catch a memory leak in our build tool. The way cosmiconfig loads '.js' config prevents it from ever getting garbage collected, even if userland code no longer needs. `require-from-string` package [creates a new `Module` record](https://github.com/floatdrop/require-from-string/blob/master/index.js#L24) which then permanently stored in memory. You can see it with this sample:

```js
const cosmiconfig = require('cosmiconfig');

const loadRC = require.resolve('cosmiconfig/lib/loadRc.js');

function loadOnce() {
    const explorer = cosmiconfig('example');
    return explorer.load(__dirname)
        .then(() => console.log(require.cache[loadRC].children.length))
}

let count = 1000;
function loop() {
    if (count > 0) {
        count--;
        loadOnce().then(loop);
    }
}

loop();
```

Provided there is a non-empty `example.config.js` nearby, you should see that `require.cache[loadRC].children.length` grows by 1 after every load and never gets cleaned up.

sudo-suhas:
### Updated dependencies

| Package                | Current | Latest | Package Type    | Reason |
|------------------------|---------|--------|-----------------|--------|
| babel-eslint           | 7.2.3   | 8.0.0  | devDependencies | No major changes, just updating babel deps to v7 [🢅](https://github.com/babel/babel-eslint/releases) |
| eslint                 | 4.6.1   | 4.7.0  | devDependencies | New features and fixes several bugs [🢅](https://eslint.org/blog/2017/09/eslint-v4.7.0-released) |
| eslint-config-prettier | 2.4.0   | 2.5.0  | devDependencies | Support for eslint-plugin-standard [🢅](https://github.com/prettier/eslint-config-prettier/blob/master/CHANGELOG.md) |
| jest                   | 20.0.4  | 21.1.0 | devDependencies | :man_shrugging: [🢅](https://github.com/facebook/jest/blob/master/CHANGELOG.md) |
| lint-staged            | 4.1.3   | 4.2.1  | devDependencies | friendlier error if config is missing [🢅](https://github.com/okonet/lint-staged/releases) |
| parse-json             | 2.2.0   | 3.0.0  | dependencies    | ES2015ify and require Node.js 4, append position to end of error message [🢅](https://github.com/sindresorhus/parse-json/compare/v2.2.0...v3.0.0) |
| prettier               | 1.6.1   | 1.7.0  | devDependencies | Some bugfixes, tweaks for JSX, TypeScript, CSS, JavaScript formatting, a couple new features.[🢅](https://github.com/prettier/prettier/releases) |
| require-from-string    | 1.2.1   | 2.0.1  | dependencies    | Cleanup `module.children` after evaluating [🢅](https://github.com/floatdrop/require-from-string/compare/v1.2.1...v2.0.0) |

`eslint-plugin-node`(`v3.0.5` → `v5.1.0`) was not updated due to peer dependency with `eslint-config-davidtheclark-node`(`"eslint-plugin-node": "^2.0.0 || ^3.0.0"`).

### Noteworthy
- Fix memory leak due to bug in `require-from-string`(`1.2.1➡2.0.1`)
- For JSON files, append position to end of error message(`2.2.0➡3.0.0`)

Closes #90

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## master
+
+- Fixed: memory leak due to bug in `require-from-string`
+- Added: for JSON files, append position to end of error message
+
 ## 3.0.0
 
 - Removed: support for loading config path using the `--config` flag. cosmiconfig will not parse command line arguments. Your application can parse command line arguments and pass them to cosmiconfig.
--- a/package.json
+++ b/package.json
@@ -71,23 +71,23 @@
   "dependencies": {
     "is-directory": "^0.3.1",
     "js-yaml": "^3.9.0",
-    "parse-json": "^2.2.0",
-    "require-from-string": "^1.1.0"
+    "parse-json": "^3.0.0",
+    "require-from-string": "^2.0.1"
   },
   "devDependencies": {
-    "babel-eslint": "^7.2.3",
+    "babel-eslint": "^8.0.0",
     "babel-plugin-transform-flow-strip-types": "^6.22.0",
-    "eslint": "^4.5.0",
+    "eslint": "^4.7.0",
     "eslint-config-davidtheclark-node": "^0.2.0",
-    "eslint-config-prettier": "^2.3.0",
+    "eslint-config-prettier": "^2.5.0",
     "eslint-plugin-flowtype": "^2.35.1",
     "eslint-plugin-node": "^3.0.5",
     "flow-bin": "^0.54.1",
     "flow-remove-types": "^1.2.1",
     "husky": "^0.14.3",
-    "jest": "^20.0.4",
-    "lint-staged": "^4.0.4",
-    "prettier": "^1.6.1"
+    "jest": "^21.1.0",
+    "lint-staged": "^4.2.1",
+    "prettier": "^1.7.0"
   },
   "engines": {
     "node": ">=4"
--- a/yarn.lock
+++ b/yarn.lock
@@ -6,6 +6,10 @@ abab@^1.0.3:
   version "1.0.3"
   resolved "https://registry.yarnpkg.com/abab/-/abab-1.0.3.tgz#b81de5f7274ec4e756d797cd834f303642724e5d"
 
+abbrev@1:
+  version "1.1.0"
+  resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.0.tgz#d0554c2256636e2f56e7c2e5ad183f859428d81f"
+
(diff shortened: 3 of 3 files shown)


sudo-suhas:
@davidtheclark What do you think of setting up greenkeeper for this repo?

davidtheclark:
@sudo-suhas: If we can configure greenkeeper to only pay attention to `dependencies` (not `devDependencies`), I'd be interested. But I don't want to get spammed about `devDependencies`.

sudo-suhas:
Alright I'll look into it.

sudo-suhas:
@davidtheclark I think we can make a patch release for this fix.

davidtheclark:
@sudo-suhas Yep, go for it.

sudo-suhas:
Since I am doing it here for the first time, please check that I have it right:

- Update CHANGELOG.md for the version about to be released and `git add CHANGELOG.md`.
- Run `npm version patch -m "Prepare %s"`
- `git push --follow-tags`
- Wait for CI to pass and then run `npm publish`

davidtheclark:
Update Changelog, yeah. The rest I guess depends on how you use Git and npm. I've always taken the more verbose route:

- Update version number in changelog and `package.json`.
- Commit the changes with a `Prepare X` message.
- Create a Git tag for the version.
- `git push && git push --tags` (Does `--follow-tags` actually do the same thing? I've tried and not had it work!)
- Then publish.

If your way is equivalent, sounds fine to me.

sudo-suhas:
I am completely okay with the more verbose 😄. 

> Does `--follow-tags` actually do the same thing? I've tried and not had it work!

See https://stackoverflow.com/questions/3745135/push-git-commits-tags-simultaneously

davidtheclark:
Oh, it looks like the reason `--follow-tags` hasn't worked for me in the past is that I haven't been using *annotated* tags, just *lightweight* tags. Didn't know about the difference. I guess I'll start using annotated tags for releases 😄

sudo-suhas:
@davidtheclark From what I can find, there is no simple way to enable greenkeeper only for [prod]dependencies. We _could_ use the method described in https://github.com/greenkeeperio/greenkeeper/issues/193. Considering we have only 4 dependencies, all of which are stable and unlikely to release major updates, I think this might be more trouble than it is worth.

davidtheclark:
@sudo-suhas Yeah, let's just stay off greenkeeper for now, then.

PR #94, opened 2017-09-25
Standardize up/down search direction wording (closes #93)

### LINKED ISSUE #93: Docs: "down the file tree" => "up the file tree"
File trees, like many others in computing (1), usually grow upside-down 😉 The root is typically referred to as the "top", and descending "down" the tree means moving into subfolders... e.g., see the wording used in http://www.linfo.org/directory_tree.html, https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/3/html/Step_by_Step_Guide/s1-navigating-cd.html, and https://www.ibm.com/support/knowledgecenter/ssw_aix_61/com.ibm.aix.cmds1/cd.htm (2).

Since cosmiconfig searches parent folders rather than subfolders for config files (_"the search continues in `./`, `../`, `../../`, `../../../`, etc."_), the conventional wording would be that it walks "_up_ the file tree" -- or even more idiomatically, "up the directory tree".

(1) Family trees sometime grow this way too (but only the descendancy kind).
(2) Did anyone else know that `cd -` toggles back and forth between two directories?!

msegado:


sudo-suhas:
@msegado Thank you for your contribution!

PR #96, opened 2017-09-29
fix: infer format based on filePath

### LINKED ISSUE #95: js config is treated as yaml
Context: https://github.com/prettier/prettier/issues/2908#issuecomment-332799662

For example, `module.exports = { foo: true }` is parsed as `{ "module.exports = { foo": "true }" }` while passing with its filename (`myConfig.js`), cosmiconfig should infer `format` based on its filename if there is no `format` provided instead of always `tryAllParsing`, which always try yaml parsing first.

something like:

https://github.com/davidtheclark/cosmiconfig/blob/ad33e75766c194eaa9f28df1d9688b3ce5a8e462/src/loadDefinedFile.js#L22

```js
const format = options.format ||
  /\.(js)$/.test(filepath)
    ? "js"
    : /\.(json)$/.test(filepath)
      ? "json"
      : /\.(yml|yaml)$/.test(filepath) ? "yaml" : undefined;

switch (format) {
```

~~I originally want to send a PR directly but failed since there are too many `expect error` tests failed I have no idea how to fix them.~~

ikatyang:
Fixes #95


src/loadDefinedFile.js
```diff
@@ -19,7 +19,16 @@ module.exports = function loadDefinedFile(
     }
 
     let parsedConfig;
-    switch (options.format) {
+
+    const format =
```
sudo-suhas: Instead of the nested ternary, could we use `path.extname`?

sudo-suhas:
Also, could you add a note to the changelog?

sudo-suhas:
@davidtheclark I'll merge this in and release it tomorrow. Let me know if you want me to hold off or you want any changes.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Fixed: infer format based on filePath
```
davidtheclark: I don't think this is a `fix`, it's a new feature, right? So a minor bump?

sudo-suhas:
@davidtheclark Just noticed that the `format` option is not documented.

PR #91, opened 2017-09-16
Setup and do changes for 100% coverage thresholds

sudo-suhas:
Code coverage went down in #87 but we did not notice it:
```
λ yarn coverage
yarn coverage v1.0.2
$ jest --coverage --coverageReporters=html --coverageReporters=text
 PASS  test\caches.test.js
 PASS  test\failed-files.test.js
 PASS  test\failed-directories.test.js
 PASS  test\successful-directories.test.js
 PASS  test\successful-files.test.js
 PASS  test\resolveDir.test.js
 PASS  test\funcRunner.test.js
 PASS  test\index.test.js

Test Suites: 8 passed, 8 total
Tests:       104 passed, 104 total
Snapshots:   0 total
Time:        2.928s
Ran all test suites.
--------------------|----------|----------|----------|----------|----------------|
File                |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
--------------------|----------|----------|----------|----------|----------------|
All files           |    98.13 |     98.6 |      100 |    97.95 |                |
 createExplorer.js  |      100 |      100 |      100 |      100 |                |
 funcRunner.js      |      100 |      100 |      100 |      100 |                |
 getDirectory.js    |      100 |      100 |      100 |      100 |                |
 index.js           |      100 |      100 |      100 |      100 |                |
 loadDefinedFile.js |    90.32 |      100 |      100 |    90.32 |       25,30,33 |
 loadJs.js          |      100 |      100 |      100 |      100 |                |
 loadPackageProp.js |      100 |      100 |      100 |      100 |                |
 loadRc.js          |    97.87 |    93.33 |      100 |    97.73 |             50 |
 parseJson.js       |      100 |      100 |      100 |      100 |                |
 readFile.js        |      100 |      100 |      100 |      100 |                |
--------------------|----------|----------|----------|----------|----------------|
Done in 6.45s.
```

This configures [coverage thresholds](https://facebook.github.io/jest/docs/en/configuration.html#coveragethreshold-object) for jest and does the necessary changes for 100% code coverage. Now, if any changes reduce the coverage, it will fail in CI.

PR #98, opened 2017-10-03
Fix UNC path issue reported in prettier/prettier-vscode#219

azz:
### DO NOT MERGE

I'm experimenting with a way to solve prettier/prettier-vscode#219, where cosmiconfig should stop moving up the directory tree in this case:

```
\\127.0.0.1\c$\a\b\c
\\127.0.0.1\c$\a\b
\\127.0.0.1\c$\a
\\127.0.0.1\c$
STOP - do not read \\127.0.0.1
```

Unfortunately, `path.dirname('\\x\y') === '\\x'`, which is not what we want.

Notes:
1. I wrote this on Mac and will need to run it through CI to see if it passes on Windows.
2. I should test this on an actual Windows machine.
3. Is it ok to have tests that only run on one platform?

### DIFF
--- a/test/util.js
+++ b/test/util.js
@@ -49,7 +49,8 @@ exports.mockStatIsDirectory = function mockStatIsDirectory(result) {
 exports.assertSearchSequence = function assertSearchSequence(
   readFileMock,
   searchPaths,
-  startCount
+  startCount,
+  isAbsolute
 ) {
   startCount = startCount || 0;
 
@@ -57,7 +58,7 @@ exports.assertSearchSequence = function assertSearchSequence(
 
   searchPaths.forEach((searchPath, idx) => {
     expect(readFileMock.mock.calls[idx + startCount][0]).toBe(
-      path.join(__dirname, searchPath)
+      path.join(isAbsolute ? '' : __dirname, searchPath)
     );
   });
 };
--- a/package.json
+++ b/package.json
@@ -72,7 +72,8 @@
     "is-directory": "^0.3.1",
     "js-yaml": "^3.9.0",
     "parse-json": "^3.0.0",
-    "require-from-string": "^2.0.1"
+    "require-from-string": "^2.0.1",
+    "unc-path-regex": "^0.1.2"
   },
   "devDependencies": {
     "babel-eslint": "^8.0.0",
--- a/src/createExplorer.js
+++ b/src/createExplorer.js
@@ -141,7 +141,8 @@ module.exports = function createExplorer(options: {
         if (result) return result;
 
         // istanbul ignore if (test only runs on Windows)
-        if (isUncStopDir(directory)) return null;
+        if (process.platform === 'win32' && isUncStopDir(directory))
+          return null;
 
         const nextDirectory = path.dirname(directory);
 
--- a/src/createExplorer.js
+++ b/src/createExplorer.js
@@ -10,6 +10,7 @@ const loadDefinedFile = require('./loadDefinedFile');
 const funcRunner = require('./funcRunner');
 const getDirectory = require('./getDirectory');
 
+// istanbul ignore next
 const isUncStopDir = (regExp => fsPath => regExp.test(fsPath))(
   new RegExp(`^${uncPathRegex().source}$`)
 );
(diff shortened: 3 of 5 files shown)


sudo-suhas:
Hey @azz, I work on a windows machine and I can help you out if you need me to test anything. Could you please help me understand the issue and the approach you are trying to fix it?

In my opinion, it's okay to have some tests that only run on windows if it can't be avoided.

Regarding `process.platform === 'win32'`, turns out detecting if the process is running on windows is [not that simple](https://github.com/jonschlinkert/is-windows/blob/63aeae70709733429c026447b7e08039b7ef0ac9/index.js#L16-L22). Could you switch it out to [`is-windows`](https://www.npmjs.com/package/is-windows) instead?

test/util.js
```diff
@@ -49,15 +49,16 @@ exports.mockStatIsDirectory = function mockStatIsDirectory(result) {
...
+  isAbsolute
 ) {
   startCount = startCount || 0;
 
   expect(readFileMock).toHaveBeenCalledTimes(searchPaths.length + startCount);
 
   searchPaths.forEach((searchPath, idx) => {
     expect(readFileMock.mock.calls[idx + startCount][0]).toBe(
-      path.join(__dirname, searchPath)
+      path.join(isAbsolute ? '' : __dirname, searchPath)
```
sudo-suhas: How about `isAbsolute ? searchPath : path.join(__dirname, searchPath)`?

azz:
I am actually super confused about this. Node should be smart enough to use `path.win32` as `require('path')`, which **does** handle UNC paths correctly. 

```js
path.win32.dirname('//machine/share$') // -> '//machine/share$'
```

So your code should actually be correct. Minus the cygwin/mingw case...

Perhaps it should be

```patch
-const nextDirectory = path.dirname(directory);
+const nextDirectory = isWindows() ? path.win32.dirname(directory) : path.dirname(directory);
```

sudo-suhas:
Just tried the example you had given:
```
λ node
> process.platform
'win32'
> path.dirname('//machine/share$')
'//machine/share$'
```

What am I missing?

azz:
Yeah, that works exactly as it should. However in prettier-vscode/issues/219, somehow cosmiconfig manages to try to read the equivalent of `//machine`

sudo-suhas:
@azz Does this have anything to do with the wrong kind of slash?

```
> path.dirname('\\127.0.0.1\c$')
'\\'
> path.dirname('//127.0.0.1/c$')
'//127.0.0.1/c$'
```

I don't know enough about network drives to know which would be the right kind of slash 😅.

I [recently changed](https://github.com/davidtheclark/cosmiconfig/pull/91/files#diff-af1dd9c8c993118830d07a83c09d4f2bL138) the logic for getting the next directory. This could fix the issue, though I am not too sure:
```
λ node
> const nextDir = (dir) => {
... const splitPath = dir.split(path.sep);
... return splitPath.length > 1 ? splitPath.slice(0, -1).join(path.sep) : null;
}
undefined
> nextDir('\\127.0.0.1\c$')
''
> path.dirname('\\127.0.0.1\c$')
'\\'
```

azz:
You need four `\`

```js
path.dirname('\\\\127.0.0.1\\c$')
```

davidtheclark:
@sudo-suhas @azz What's the status of this PR?

sudo-suhas:
Not able to replicate the issue(see https://github.com/davidtheclark/cosmiconfig/pull/98#issuecomment-333835175) therefore not sure about merging it either.

davidtheclark:
@sudo-suhas Well, I'd be fine with closing it as stale and waiting to see if the issue arises again. I'll leave it up to you?

azz:
Yeah I ended up out of my depth on this one, found it difficult to replicate and test. It's a fairly minor issue so we can wait to see if it is reported again.

GeekMode:
Found the error. It's actually in the 'resolve' module's nodeModulesPaths function (filename: node-modules-paths.js).

On line 40, they use path.join which removes the initial backslash and resulting in '\\\\IP_ADDRESS\SOME_LOCATION' being turned into '\\IP_ADDRESS\SOME_LOCATION'.

A simple switch from `path.join` to `path.resolve` should correct this issue in all cases I can think of.

PR #104, opened 2018-01-12
Start search from process.cwd()

### LINKED ISSUE #102: Docs misunderstanding
Hi!

Docs says `Starting from process.cwd() (or some other directory defined by the searchPath argument to load())`, but the code (https://github.com/davidtheclark/cosmiconfig/blob/master/src/createExplorer.js#L98) actually does nothing at all if a search path is not provided.

From my understanding, it should use `process.cwd()` by default if no path is provided.

sudo-suhas:
~~This adds an option `startDir`  with the defaults as `process.cwd()` and is used as the start directory for search if no parameters are passed to `explorer.load`.~~

This sets default value for `searchPath` as `process.cwd()` in `explorer.load`.

Closes #102.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ var cosmiconfig = require('cosmiconfig');
 
 var explorer = cosmiconfig(yourModuleName[, options]);
 
-explorer.load(yourSearchPath)
+explorer.load()
   .then((result) => {
     // result.config is the parsed configuration object
     // result.filepath is the path to the config file that was found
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -143,6 +143,13 @@ Instead of *just* looking for `.goldengrahamsrc` (no extension), it will also lo
 - `.goldengrahamsrc.yml`
 - `.goldengrahamsrc.js`
 
+##### startDir
```
davidtheclark: Why do you think it's preferable to add this new option instead of having the `searchPath` argument default to `process.cwd()`?

sudo-suhas: Mostly for symmetry with `configPath`. Can change it if you feel that'd be better.

davidtheclark: Yeah, I think a default value for `searchPath` makes just as much sense here while avoiding an addition to the API. Unless you feel differently, I'd prefer that approach.

sudo-suhas: Agreed and done.

Issue #106, opened 2018-02-04
Would you take a PR to throw an error for all empty configs?


olsonpm:
Current behavior:  

When using `load(null, configPath)` [cosmiconfig throws an error](https://github.com/davidtheclark/cosmiconfig/blob/master/src/loadDefinedFile.js#L18-L20)

When using `load(dirPath)` [cosmiconfig continues to traverse up the directory path](https://github.com/davidtheclark/cosmiconfig/blob/master/src/loadRc.js#L35)

If this was intentional then please help me understand why.  Otherwise I would like to create a PR to throw in all cases.  I don't really care whether an error is thrown, ~~we're mostly looking for a way to differentiate empty configurations from those which are non-existent.  e.g. `package.json` with an empty `prettier` property should behave differently from a package.json without a `prettier` property.~~

Edit: I was basing the above off of issue comments I was reading in the prettier repo - not off of local tests.  When cosmiconfig sees a package.json file with `"prettier": {}`, it correctly returns an empty object and does not traverse up the file tree.  However it still treats an empty *rc file as non-existent.  I now understand this is an edge case that really can really only help expose configuration errors made by developers.
<br>

#### Why
`prettier-atom` would like to allow users to specify settings within the plugin for one-off cases where files do not belong to projects.  However if **any configuration is found, even empty** then we would like to use that configuration instead.  Empty signifying default settings.

sudo-suhas:
@olsonpm, When loading a specified file, the behavior was inconsistent if the file was empty(see https://github.com/davidtheclark/cosmiconfig/pull/78#issuecomment-318810892) and this was fixed in `v3.0.0`. However, when performing a search up the directory tree, I am not entirely sure that we should throw an error when we find an empty config file. 

> `package.json` with an empty prettier property should behave differently from a `package.json` without a prettier property.

If the pkg has an empty property, the search resolves with `{}`. The same is true for a JSON file with `{}` as the contents. So would it be enough to check if the value that `cosmiconfig` search resolves to, is an empty object?

<details>
<summary>resolve to empty cfg</summary>

**`package.json`:**

```json
{
  "name": "prettier-pkg-empty-cfg",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "prettier": "^1.10.2"
  },
[...]

PR #107, opened 2018-02-08
Treat empty files differently from non-existent

olsonpm:
- Return an 'undefined' config along with the file path for empty
  configuration files when both traversing directories and loading
  defined files.  Previously empty configuration files would prompt
  an error upon loading defined files and be silently passed over
  during directory traversal.

BREAKING CHANGE: treats empty files differently

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,26 +192,38 @@ In the event that the file does not have an extension or the extension is unreco
 
 ### Instance methods (on `explorer`)
 
-#### `search([searchPath])`
+#### `search([searchPath][, searchOptions])`
 
 Searches for a configuration file. Returns a Promise that resolves with `null`, if nothing is found, or an object with two properties:
 - `config`: The loaded and parsed configuration.
 - `filepath`: The filepath where this configuration was found.
 
+When the [search option](#searchoptions) `ignoreEmpty` is set to false and an empty configuration file is found, the result will have `config` set to `undefined` and an additional property `isEmpty` set to true.
+
 ```js
 explorer.search() // searches from process.cwd()
 explorer.search('start/search/here');
 explorer.search('start/search/at/this/file.css');
+
+// will stop on an empty configuration file
+explorer.search('start/search/here', { ignoreEmpty: false })
 ```
 
 If you provide `searchPath`, cosmiconfig will start its search at `searchPath` and continue to search up the directory tree, as documented above.
 By default, `searchPath` is set to `process.cwd()`.
+
+##### searchOptions
+
+Can be an object containing a single property `ignoreEmpty`.  This defaults to true and tells cosmiconfig whether to continue searching
+for files when an empty one is found.
 
 #### `load([configPath])`
 
 Loads a configuration file directly.  Rejects with an error if the file does not exist, if the file contents are empty, or if the file cannot be parsed.  Upon success it returns a Promise that resolves with an object with two properties:
 - `config`: The loaded and parsed configuration.
 - `filepath`: The filepath where this configuration was found.
+
+If the file is empty then `config` will be set to `undefined` and an additional property `isEmpty` will be set to true
 
 ```js
 explorer.load() // loads options.configPath
--- a/src/createExplorer.js
+++ b/src/createExplorer.js
@@ -51,16 +51,74 @@ module.exports = function createExplorer(options: {
   }
 
   function search(
-    searchPath: string
+    searchPath: string,
+    searchOptions: { ignoreEmpty?: boolean }
   ): Promise<?cosmiconfig$Result> | ?cosmiconfig$Result {
+    const sanitizedOpts = Object.assign(
+      {},
+      {
+        ignoreEmpty: true,
+      },
+      searchOptions
+    );
+
     if (!searchPath) searchPath = process.cwd();
 
     const absoluteSearchPath = path.resolve(process.cwd(), searchPath);
     const searchPathDir = getDirectory(absoluteSearchPath, options.sync);
 
     return searchPathDir instanceof Promise
-      ? searchPathDir.then(searchDirectory)
-      : searchDirectory(searchPathDir);
+      ? searchPathDir.then(pathDir => searchDirectory(pathDir, sanitizedOpts))
+      : searchDirectory(searchPathDir, sanitizedOpts);
+  }
+
+  function searchDirectory(
+    directory: string,
+    searchOptions: { ignoreEmpty: boolean }
+  ): Promise<?cosmiconfig$Result> | ?cosmiconfig$Result {
+    if (searchCache && searchCache.has(directory)) {
+      return searchCache.get(directory);
+    }
+
+    const result = funcRunner(!options.sync ? Promise.resolve() : undefined, [
+      () => {
+        if (!packageProp) return;
+        return loadPackageProp(directory, {
+          packageProp,
+          sync: options.sync,
+        });
+      },
+      result => {
+        if (result || !options.rc) return result;
+        return loadRc(path.join(directory, options.rc), {
+          ignoreEmpty: searchOptions.ignoreEmpty,
+          sync: options.sync,
+          rcStrictJson: options.rcStrictJson,
+          rcExtensions: options.rcExtensions,
+        });
+      },
+      result => {
+        if (result || !options.js) return result;
+        return loadJs(path.join(directory, options.js), {
+          ignoreEmpty: searchOptions.ignoreEmpty,
+          sync: options.sync,
+        });
+      },
+      result => {
+        if (result) return result;
+
+        const nextDirectory = path.dirname(directory);
+
+        if (nextDirectory === directory || directory === options.stopDir)
+          return null;
+
+        return searchDirectory(nextDirectory, searchOptions);
+      },
+      transform,
+    ]);
+
+    if (searchCache) searchCache.set(directory, result);
+    return result;
   }
 
   function load(
@@ -111,50 +169,6 @@ module.exports = function createExplorer(options: {
         ? loadResult.then(transform)
         : transform(loadResult);
     if (loadCache) loadCache.set(absoluteConfigPath, result);
-    return result;
-  }
-
-  function searchDirectory(
-    directory: string
-  ): Promise<?cosmiconfig$Result> | ?cosmiconfig$Result {
-    if (searchCache && searchCache.has(directory)) {
-      return searchCache.get(directory);
-    }
-
-    const result = funcRunner(!options.sync ? Promise.resolve() : undefined, [
-      () => {
-        if (!packageProp) return;
-        return loadPackageProp(directory, {
-          packageProp,
-          sync: options.sync,
-        });
-      },
-      result => {
-        if (result || !options.rc) return result;
-        return loadRc(path.join(directory, options.rc), {
-          sync: options.sync,
-          rcStrictJson: options.rcStrictJson,
-          rcExtensions: options.rcExtensions,
-        });
-      },
-      result => {
-        if (result || !options.js) return result;
-        return loadJs(path.join(directory, options.js), { sync: options.sync });
-      },
-      result => {
-        if (result) return result;
-
-        const nextDirectory = path.dirname(directory);
-
-        if (nextDirectory === directory || directory === options.stopDir)
-          return null;
-
-        return searchDirectory(nextDirectory);
-      },
-      transform,
-    ]);
-
-    if (searchCache) searchCache.set(directory, result);
     return result;
   }
 
--- a/src/createParseFile.js
+++ b/src/createParseFile.js
@@ -1,7 +1,11 @@
 'use strict';
 
-module.exports = function createParseFile(filepath, parse, ignoreEmpty) {
-  return function parseFile(content) {
+module.exports = function createParseFile(
+  filepath: string,
+  parse: (string, string) => Object,
+  ignoreEmpty: boolean
+): string => cosmiconfig$Result {
+  return function parseFile(content: string): cosmiconfig$Result {
     const isEmpty = content === '';
     if (content == null || (isEmpty && ignoreEmpty)) return null;
 
--- a/src/createParseFile.js
+++ b/src/createParseFile.js
@@ -1,0 +1,12 @@
+'use strict';
+
+module.exports = function createParseFile(filepath, parse, ignoreEmpty) {
+  return function parseFile(content) {
+    const isEmpty = content === '';
+    if (content == null || (isEmpty && ignoreEmpty)) return null;
+
+    return isEmpty
+      ? { config: undefined, filepath, isEmpty }
+      : { config: parse(content, filepath), filepath };
+  };
+};
--- a/src/loadDefinedFile.js
+++ b/src/loadDefinedFile.js
@@ -16,7 +16,7 @@ module.exports = function loadDefinedFile(
 ): Promise<?cosmiconfig$Result> | ?cosmiconfig$Result {
   function parseContent(content: ?string): ?cosmiconfig$Result {
     if (!content) {
-      throw new Error(`Config file is empty! Filepath - "${filepath}".`);
+      return { config: undefined, filepath, isEmpty: true };
     }
 
     let parsedConfig;
--- a/src/loadJs.js
+++ b/src/loadJs.js
@@ -3,19 +3,17 @@
 
 const requireFromString = require('require-from-string');
 const readFile = require('./readFile');
+const createParseFile = require('./createParseFile');
 
 module.exports = function loadJs(
   filepath: string,
-  options: { sync?: boolean }
+  options: { ignoreEmpty: boolean, sync?: boolean }
 ): Promise<?cosmiconfig$Result> | ?cosmiconfig$Result {
-  function parseJsFile(content: ?string): ?cosmiconfig$Result {
-    if (!content) return null;
-
-    return {
-      config: requireFromString(content, filepath),
-      filepath,
-    };
-  }
+  const parseJsFile = createParseFile(
+    filepath,
+    requireFromString,
+    options.ignoreEmpty
+  );
 
   return !options.sync
     ? readFile(filepath).then(parseJsFile)
--- a/src/loadRc.js
+++ b/src/loadRc.js
@@ -6,15 +6,25 @@ const requireFromString = require('require-from-string');
 const readFile = require('./readFile');
 const parseJson = require('./parseJson');
 const funcRunner = require('./funcRunner');
+const createParseFile = require('./createParseFile');
 
 module.exports = function loadRc(
   filepath: string,
   options: {
+    ignoreEmpty: boolean,
     sync?: boolean,
     rcStrictJson?: boolean,
     rcExtensions?: boolean,
   }
 ): Promise<?cosmiconfig$Result> | ?cosmiconfig$Result {
+  const parseYaml = (content, filename) => yaml.safeLoad(content, { filename });
+  const parse = options.rcStrictJson ? parseJson : parseYaml;
+  const parseExtensionlessRcFile = createParseFile(
+    filepath,
+    parse,
+    options.ignoreEmpty
+  );
+
   if (!options.sync) {
     return readFile(filepath)
       .then(parseExtensionlessRcFile)
@@ -31,73 +41,32 @@ module.exports = function loadRc(
     return null;
   }
 
-  function parseExtensionlessRcFile(content: ?string): ?cosmiconfig$Result {
-    if (!content) return null;
-    const pasedConfig = options.rcStrictJson
-      ? parseJson(content, filepath)
-      : yaml.safeLoad(content, { filename: filepath });
-    return {
-      config: pasedConfig,
-      filepath,
-    };
-  }
-
   function loadRcWithExtensions() {
-    let foundConfig = null;
-    return funcRunner(readRcFile('json'), [
-      (jsonContent: ?string) => {
-        // Since this is the first try, config cannot have been found, so don't
-        // check `if (foundConfig)`.
-        if (jsonContent) {
-          const successFilepath = `${filepath}.json`;
-          foundConfig = {
-            config: parseJson(jsonContent, successFilepath),
-            filepath: successFilepath,
-          };
-        } else {
-          return readRcFile('yaml');
-        }
-      },
-      (yamlContent: ?string) => {
-        if (foundConfig) {
-          return;
-        } else if (yamlContent) {
-          const successFilepath = `${filepath}.yaml`;
-          foundConfig = {
-            config: yaml.safeLoad(yamlContent, { filename: successFilepath }),
-            filepath: successFilepath,
-          };
-        } else {
-          return readRcFile('yml');
-        }
-      },
-      (ymlContent: ?string) => {
-        if (foundConfig) {
-          return;
-        } else if (ymlContent) {
-          const successFilepath = `${filepath}.yml`;
-          foundConfig = {
-            config: yaml.safeLoad(ymlContent, { filename: successFilepath }),
-            filepath: successFilepath,
-          };
-        } else {
-          return readRcFile('js');
-        }
-      },
-      (jsContent: ?string) => {
-        if (foundConfig) {
-          return;
-        } else if (jsContent) {
-          const successFilepath = `${filepath}.js`;
-          foundConfig = {
-            config: requireFromString(jsContent, successFilepath),
-            filepath: successFilepath,
-          };
-        } else {
-          return;
-        }
-      },
-      () => foundConfig,
+    function loadExtension(
+      extension: string,
+      parse: (content: string, filename: string) => Object
+    ) {
+      const fpath = `${filepath}.${extension}`;
+      const parseRcFile = createParseFile(fpath, parse, options.ignoreEmpty);
+
+      // Check the result from the previous `loadExtension` invocation. If result
+      // isn't null, just return that.
+      return result => {
+        if (result != null) return result;
+
+        // Try to load the rc file for the given extension.
+        return funcRunner(readRcFile(extension), [parseRcFile]);
+      };
+    }
+
+    const parseYml = (content: string, filename: string) =>
+      yaml.safeLoad(content, { filename });
+
+    return funcRunner(!options.sync ? Promise.resolve() : undefined, [
+      loadExtension('json', parseJson),
+      loadExtension('yaml', parseYml),
+      loadExtension('yml', parseYml),
+      loadExtension('js', requireFromString),
     ]);
   }
 
--- a/src/loadRc.js
+++ b/src/loadRc.js
@@ -55,7 +55,7 @@ module.exports = function loadRc(
         if (result != null) return result;
 
         // Try to load the rc file for the given extension.
-        return funcRunner(readRcFile(extension), [parseRcFile]);
+        return funcRunner(readRcFile(fpath), [parseRcFile]);
       };
     }
 
@@ -70,10 +70,7 @@ module.exports = function loadRc(
     ]);
   }
 
-  function readRcFile(extension: string): Promise<?string> | ?string {
-    const filepathWithExtension = `${filepath}.${extension}`;
-    return !options.sync
-      ? readFile(filepathWithExtension)
-      : readFile.sync(filepathWithExtension);
+  function readRcFile(filepath: string): Promise<?string> | ?string {
+    return !options.sync ? readFile(filepath) : readFile.sync(filepath);
   }
 };
--- a/test/failed-directories.test.js
+++ b/test/failed-directories.test.js
@@ -347,6 +347,7 @@ describe('cosmiconfig', () => {
               throw { code: 'ENOENT' };
             case absolutePath('a/b/c/d/e/f/.foorc.json'):
             case absolutePath('a/b/c/d/e/f/.foorc.yaml'):
+              // This proves the default 'ignoreEmpty' skips over these files as expected
               return '';
             case absolutePath('a/b/c/d/e/f/.foorc.yml'):
               return 'found: thing: true';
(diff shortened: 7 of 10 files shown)


olsonpm:
Something concrete to discuss for #106

I assume all "empty config" scenarios are error cases, so I could also see us throwing an error and attaching a property 'filepath' to it so consumers (e.g. prettier-atom) can pass that debugging information onto users.

Just let me know how you guys prefer to handle empty configs - but I do feel they should be somehow treated differently from non-existent files.  I also feel it should be the same whether the config is loaded directly (defined file) or via directory traversal.

src/loadRc.js
```diff
@@ -32,12 +32,18 @@ module.exports = function loadRc(
...
 
   function parseExtensionlessRcFile(content: ?string): ?cosmiconfig$Result {
-    if (!content) return null;
-    const pasedConfig = options.rcStrictJson
+    if (typeof content !== 'string') return null;
+    if (content === '')
+      return {
+        config: undefined,
+        filepath,
+      };
```
davidtheclark: Looks like Prettier probably made this uglier than you intended. Could you add some `{}` around the block?

davidtheclark:
I like this change. @sudo-suhas, do you?

> I assume all "empty config" scenarios are error cases, so I could also see us throwing an error and attaching a property 'filepath' to it so consumers (e.g. prettier-atom) can pass that debugging information onto users.

I'm thinking I'd rather leave that to the consumer. They can decide how to validate their config files. As long as they can reliably identify an empty file (which this PR makes possible), I'm not sure an error is necessary.

> I do feel they should be somehow treated differently from non-existent files

I agree. @sudo-suhas @olsonpm are we confident this change covers all the cases @sudo-suhas laid out here in a consistent way: https://github.com/davidtheclark/cosmiconfig/pull/78#issuecomment-318810892?

> Just let me know how you guys prefer to handle empty configs - but I do feel they should be somehow treated differently from non-existent files. I also feel it should be the same whether the config is loaded directly (defined file) or via directory traversal.

I think my preference is to do what this PR does. @olsonpm what do you think are the open questions?

src/loadRc.js
olsonpm: Ah sorry about that.  Will do

olsonpm:
> I'm thinking I'd rather leave that to the consumer.

Makes sense.  Thanks
<br>

> are we confident this change covers all the cases sudo-suhas laid out here in a consistent way?

I'll test that and edit this post with my results
<br>

> what do you think are the open questions?

Nothing I can think of.  I would like sudo-suhas' approval before I manually test this branch (I'm never confident with first PRs, and I don't like testing before the implementation is agreed-upon).  After I test I'll remove the `WIP`

sudo-suhas:
I was out of town during the last few days and hence the delayed response. 

My thoughts:
- Would it be better to accept a config option to determine behavior for empty config files? Might be a little complicated to implement as it would need to handle the cases for `'resolveEmpty'|'ignoreEmpty'|'throwEmpty'`. This would simplify scenarios like 'load only a non-empty cfg'.
- Instead of returning `{ config: undefined, filepath }`, how about adding a boolean field `emptyCfg` instead? I think this might be cleaner. Would also avoid an extreme edge case of JS file with `module.exports = undefined`

src/loadJs.js
```diff
@@ -9,10 +9,13 @@ module.exports = function loadJs(
   options: { sync?: boolean }
 ): Promise<?cosmiconfig$Result> | ?cosmiconfig$Result {
   function parseJsFile(content: ?string): ?cosmiconfig$Result {
-    if (!content) return null;
+    if (typeof content !== 'string') return null;
```
sudo-suhas: How about changing this check to `content == null`? I think that is clearer in terms of intention.

olsonpm: hmm, i thought the linter would get angry about double equals but i'll make that change if not

olsonpm:
i'm still apathetic as to the implementation and will implement whatever you guys agree on

olsonpm:
haha okay

> Would it be better to accept a config option to determine behavior for empty config files

I think until a consumer has a need for empty config files other than for friendly error messages, we should hold off on allowing it to be customized.
<br>

> Instead of returning { config: undefined, filepath }, how about adding a boolean field emptyCfg instead?

cleaner indeed 👍

sudo-suhas:
> I think until a consumer has a need for empty config files other than for friendly error messages, we should hold off on allowing it to be customized.

I was thinking of the scenario the other way around. Currently, `cosmiconfig` will resolve with the first non-empty cfg it finds while searching up the directory tree. But with this change, to get the old behavior, the search would need to be restarted from the directory with the empty file.

olsonpm:
In that case would `ignoreEmpty` only apply for `search` then?  (using the new terminology search/load)

sudo-suhas:
I think, if the option is part of the config used to instantiate the explorer instance, it should be applicable for both. Otherwise, we could accept it as the second parameter with `ignoreEmpty` as the default for `search` and `throwEmpty` for `load`.

olsonpm:
Yeah I'm torn.  The argument for backward compatibility is good for sure, but I don't think empty file behavior should differ between `search` and `load` meaning I would prefer that option to exist for both.  But then I think ignoring an empty file when `load`ing it directly would be extremely confusing behavior.  So in my opinion the decision comes down to a value judgement between backward compatibility and simplicity.  I don't want to make that call haha.

davidtheclark:
Ok, to help finalize a decision — here's what I see as the use cases:

- Using `load`, if the config I point to is empty I want an error.
- Using `load`, if the config I point to is empty I *don't* want an error; just give me `undefined` for the content.
- Using `search`, if an empty config is found I want an error.
- Using `search`, if an empty config is found I *don't* want an error, but I do want the search to stop; just give me `undefined` for the content.
- Using `search`, if an empty config is found I want the search to skip over it and continue.

For throwing vs not throwing: I'm in favor of leaving this in the hands of the user. cosmiconfig can just give them `undefined` content, and the user can determine how to respond. I don't think an edge case like `module.exports = undefined` is worth worrying about, as that seems to indicate the exact same intention as a completely empty `yaml` file, for example.

The idea of skipping over an empty config only applies to `search`, doesn't really have any parallel for `load`. If we follow through with splitting out `search` and `load` functions, the second argument to either could an options argument, and `ignoreEmpty` could be an option for `search`, not something you an option for `cosmiconfig`, when you instantiate the explorer (`load` wouldn't have any options yet).

How does this sound @sudo-suhas @olsonpm?

sudo-suhas:
It sounds good @davidtheclark. Just to summarise,

- If `load` gets an empty file, return `{ config: undefined, filepath, isEmpty: true }`. The user can choose to throw an error for this by checking the result.
- `search` uses the `ignoreEmpty` parameter to either return the empty config result or continue the search.

Do you think it would make sense to recommend throwing the error in the transform function? We could implement and export this so the user could do something like this:

```js
const explorer = cosmiconfig(moduleName, {
  transform: cosmiconfig.throwOnEmpty,
})
```

davidtheclark:
@sudo-suhas I'm not sure I'm convinced that a `isEmpty: true` in the return object or a `throwOnEmpty` helper are worth the additional API surface for this module.

For `isEmpty: true` — do you really think there will be a user who needs to distinguish between a JS file that's completely empty and one that's `module.exports = undefined`? That seems like an unrealistic edge case, or at least one that we don't need to cater to.

For `throwOnEmpty` — I don't think this provides enough convenience that it's worth an API addition. When the user gets a config object, they can check if the content is `undefined` and respond accordingly. 

Do you think I'm not understanding some benefits?

sudo-suhas:
> For `isEmpty`: true — do you really think there will be a user who needs to distinguish between a JS file that's completely empty and one that's module.exports = undefined? That seems like an unrealistic edge case, or at least one that we don't need to cater to.

No, I am in agreement with you on this. My intention was not to say that this is the main reason for the flag. To clarify, these are the reasons I am advocating it

- When we return `{ config: undefined, filepath }` as the result, we are attaching special meaning to the `undefined` which might not be immediately obvious.
- I think having a `isEmpty` flag is explicit. In the client code, for someone reading it without complete familiarity with `cosmiconfig` API, consider the following:
  ```js
  explorer.search()
    .then(result => {
      // Usage with proposed
      if (result !== null && result.isEmpty) { /* do something */ }

      // Without the flag present
      if (result !== null && result.config === undefined) { /* do something */ }
    })
  
  ```
  In my opinion the proposed syntax is much easier to understand without having to consult the documentation to understand what `undefined` signifies.

> For `throwOnEmpty` — I don't think this provides enough convenience that it's worth an API addition. When the user gets a config object, they can check if the content is `undefined` and respond accordingly.

Alright, fair enough. How about just having the following in the documentation:

```js
// To throw an error on getting an empty configuration file
const explorer = cosmiconfig(moduleName, {
  transform: result => {
    if (result.isEmpty) {
      throw new Error(`Config file is empty! Filepath - "${result.filepath}".`);
    }

    return result;
  },
})
```

olsonpm:
yeah to me using `isEmpty` is just like avoiding a `-1` magic number return which is then documented to mean something.  Where every time you see `-1` in the code you have to either remember what it means or refer to the documentation, and my memory for things like that is nil.

davidtheclark:
Ok, I'm fine with having `isEmpty` since you both want it 👍

@olsonpm — I think we're all agreed upon the goals now, right?

olsonpm:
I think I addressed

- [use double equals null](https://github.com/davidtheclark/cosmiconfig/pull/107#discussion_r167470137) instead of checking for `typeof`
- fixed issue where the config wasn't being set to undefined in loadRc (previously I had missed that portion)

- added `isEmpty: true` only when the content is empty.
  
  *I decided against setting `isEmpty` to false explicitly because it seemed unnecessary.  I also made sure isEmpty wasn't being set to false anywhere to prevent users from checking `isEmpty === false`*

- I decided against documenting the error on `isEmpty` transform because
  1. The transform portion of the documentation was already fairly large
  2. Adding an error check goes against the previous recommendation of utilizing the transform for extra io whose result can be cached.

  Regardless, I'll add it if you want me to.

Finally, I kept the code verbose in the new tests for empty rc files because that seemed to be consistent with the existing code.  Let me know if you want that slimmed down.

olsonpm:
hmm re-reading this thread now I didn't realize you guys agreed on adding `ignoreEmpty`.  Also I'll rebase this onto the search/load branch, knowing I'll have to rebase it on master once it gets mereged.

src/loadRc.js
```diff
@@ -32,26 +32,44 @@ module.exports = function loadRc(
...
     return {
-      config: pasedConfig,
+      config: parsedConfig,
       filepath,
     };
   }
 
   function loadRcWithExtensions() {
-    let foundConfig = null;
+    let foundConfig: ?cosmiconfig$Result = null;
```
sudo-suhas: I had wanted to refactor this some time back to avoid the code repetition. @olsonpm, @davidtheclark, what do you think of this implementation for `loadRcWithExtensions`:

```js
function loadRcWithExtensions() {
  function loadExtn(
    extn: string,
    parse: (content: string, filename: string) => Object
  ) {
    // Check the result from the previous `loadExtn` invocation. If result
    // isn't null, just return that.
    return result => {
      if (result != null) return result;

      // Try to load the rc file for the given extension.
      return funcRunner(readRcFile(extn), [
        content => {
          if (content == null) return;

          const fp = `${filepath}.${extn}`;
          return content === ''
            ? { config: undefined, filepath: fp, isEmpty: true }
            : { config: parse(content, fp), filepath: fp };
        },
      ]);
    };
  }

  const parseYml = (content: string, filename: string) =>
    yaml.safeLoad(content, { filename });

  return funcRunner(!options.sync ? Promise.resolve() : undefined, [
    loadExtn('json', parseJson),
    loadExtn('yaml', parseYml),
    loadExtn('yml', parseYml),
    loadExtn('js', requireFromString),
  ]);
}
```

src/loadJs.js
```diff
@@ -6,15 +6,20 @@ const readFile = require('./readFile');
...
+  options: { ignoreEmpty: boolean, sync?: boolean }
 ): Promise<?cosmiconfig$Result> | ?cosmiconfig$Result {
   function parseJsFile(content: ?string): ?cosmiconfig$Result {
-    if (!content) return null;
+    const isEmpty = content === '';
+    if (content == null || (isEmpty && options.ignoreEmpty)) return null;
 
-    return {
-      config: requireFromString(content, filepath),
+    const config = isEmpty ? undefined : requireFromString(content, filepath);
```
sudo-suhas: I think this section could be improved a little. How about this:

```js
return isEmpty
  ? { config: undefined, filepath, isEmpty: true }
  : { config: requireFromString(content, filepath), filepath };
```

sudo-suhas:
I removed one review comment which was no longer relevant.

src/loadJs.js
olsonpm: just pushed

src/loadRc.js
olsonpm: I think I remember reading this conversation you guys had.  like you, I tend to go overboard on removing code repetition so davidtheclark would be better suited to say one way or the other.  I agree it's good to revisit with the amount of repeated code I added.

davidtheclark: Looks good. My only suggestion is spelling out `extension` instead of abbreviating it 😄

davidtheclark:
@olsonpm thanks for the continued excellent work! Would you like to be added as a collaborator to this project, on GitHub and npm?

Once you and @sudo-suhas are pleased with this let's merge. Then please give me a little while so I can find some time to take a pass myself over the code and especially the documentation before we release 5.0.

src/loadRc.js
sudo-suhas: That's side effects of [working on Go :wink:](https://github.com/golang/go/wiki/CodeReviewComments#variable-names)

olsonpm: pushed

olsonpm:
- Added a couple notes to the changelog
- Slimmed down `src/loadRc.js`

@sudo-suhas don't worry about nit-picks if you have any.  I'd much prefer to write code that fits your guys' "voice" per se.

@davidtheclark I'd like to be a collaborator.  Thanks.  I haven't done much in terms of managing npm packages so if something happened I'd be learning how to fix it rather than just fixing it haha.  Any code modifications I'd also still be looking heavily for your and suhas' input because I'm not confident in writing code that matches your style.  But I'm willing to write code and definitely help out with issues.

Finally - no rush on getting it merged.

sudo-suhas:
The description you are giving fits @davidtheclark rather than me 😃. I am kinda new here too.

olsonpm:
Guess it also depends on whether it makes sense for the options to exist in the explorer as a default, specified per-call, or both.  I haven't used cosmiconfig enough to have a feel for what scope is appropriate.

sudo-suhas:
There is also the option of accepting `searchOptions` and `loadOptions` in the `createExplorer` options but that might be taking it too far.

olsonpm:
yeah that was my thought as well.  Maybe wait until someone requests it?

Another thought I had was to make load and search take a single argument each, to get rid of the awkward `search(undefined, { ignoreEmpty: false })`

e.g. 

```js
// instead of this
function search(searchPath, searchOptions = {}) {
  const { ignoreEmpty = true } = searchOptions
  // ...
}

// we could do this
function search(searchOptions = {}) {
  const {
    ignoreEmpty = true,
    searchPath = process.cwd()
  } = searchOptions
  // ...
}
```
<br>

I know the more common approach in the javascript wild is to collapse the arguments, e.g.

```js
function search(searchPath, searchOptions) {
  if (typeof searchPath === 'object') {
    searchOptions = searchPath
  }
  // ...
}
```

I personally think this approach makes input difficult to validate so I don't use it in my own code, but it's another option.  Again, probably an unnecessary problem to tackle atm

sudo-suhas:
> Another thought I had was to make load and search take a single argument each

I like this idea and am in agreement with you w.r.t collapsing the arguments.

test/failed-directories.test.js
```diff
@@ -319,14 +387,146 @@ describe('cosmiconfig', () => {
...
+        };
+        expect(search({ ignoreEmpty: false, sync: true })).toEqual(
+          expectedResult
+        );
+        return expect(
+          search({ ignoreEmpty: false, sync: false })
+        ).resolves.toEqual(expectedResult);
+      });
+
+      it('returns an undefined config, file path, and isEmpty = true for empty yml', () => {
```
sudo-suhas: I don't think this is required. Testing for `yaml` should do. Otherwise maybe we should look into integrating [`jest-in-case`](https://github.com/atlassian/jest-in-case).

test/failed-directories.test.js
```diff
@@ -124,6 +124,36 @@ describe('cosmiconfig', () => {
       ]);
```
sudo-suhas: Could you add a couple of tests for failed search with the default behavior of `ignoreEmpty`?

test/failed-directories.test.js
```diff
@@ -124,6 +124,36 @@ describe('cosmiconfig', () => {
...
     });
 
+    it('returns an undefined config, file path, and isEmpty = true for empty rc file', () => {
```
sudo-suhas: Instead of `undefined config, file path, and isEmpty = true`, how about `empty config result`?

olsonpm:
sounds good.

README.md
```diff
@@ -192,27 +192,39 @@ In the event that the file does not have an extension or the extension is unreco
...
 explorer.search('start/search/at/this/file.css');
+
+// will stop on an empty configuration file
+explorer.search('start/search/here', { ignoreEmpty: false })
 ```
 
 If you provide `searchPath`, cosmiconfig will start its search at `searchPath` and continue to search up the directory tree, as documented above.
 By default, `searchPath` is set to `process.cwd()`.
 
+##### searchOptions
```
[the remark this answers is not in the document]
olsonpm: I used the header to link from above

> When the [search option](#searchoptions) `ignoreEmpty` is set to false

I did this to ensure we weren't just dropping a term that hasn't been introduced yet, and thought it would be odd to place the definition above the paragraph where it was linked.  I'm open to a reorganization though

src/loadRc.js
```diff
@@ -32,72 +33,61 @@ module.exports = function loadRc(
   }
 
   function parseExtensionlessRcFile(content: ?string): ?cosmiconfig$Result {
-    if (!content) return null;
-    const pasedConfig = options.rcStrictJson
+    const isEmpty = content === '';
```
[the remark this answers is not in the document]
olsonpm: Really all three methods are doing the exact same thing just parsing differently.  Instead of calling a helper you could utilize a factory e.g.

*not sure 'factory' is the correct term here, I'm no academic ha*

```js
// createParseFile.js
module.exports = function createParseFile(filepath, parse, ignoreEmpty) {
  return function parseFile(content) {
    const isEmpty = content === '';
    if (content == null || (isEmpty && ignoreEmpty)) return null;

    if (isEmpty) {
      return {
        config: undefined,
        filepath,
        isEmpty,
      };
    }
    
    return {
      config: parse(parsedConfig),
      filepath,
    };
  }
}
```

and consume it like

```js
// loadRc.js
const parseYaml = (content, filepath) => yaml.safeLoad(content, { filename: filepath });

const parse = options.rcStrictJson
  ? parseJson
  : parseYaml

const parseExtensionlessRcFile = createParseFile(filepath, parse, ignoreEmpty)

// ...
// inside loadExtension

const fpath = `${filepath}.${extn}`
const parseRcFile = createParseFile(fpath, parse, ignoreEmpty)

// then replace
return funcRunner(readRcFile(extn), [
// with
return funcRunner(readRcFile(extn), [parseRcFile])
```

```js
// loadJs.js
const parseJsFile = createParseFile(filepath, requireFromString, ignoreEmpty)
```

test/failed-directories.test.js
olsonpm: I think prior to shortening that portion of the code, it was necessary for 100% coverage.  You're right it can be removed now, but so can the empty `js` and `yml`.  Should we remove those as well?

test/failed-directories.test.js
olsonpm: do you think we can just utilize the existing failure tests and toss an empty file in them?  Thanks for catching this btw

olsonpm:
Alright everything I haven't commented on should be fixed/pushed.  Thanks much for taking a look.

README.md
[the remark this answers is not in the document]
sudo-suhas: Okay, that makes sense. Just removing the bulletting and adding a newline after the heading should do.

README.md
```diff
@@ -192,27 +192,39 @@ In the event that the file does not have an extension or the extension is unreco
...
+- Can be an object containing a single property `ignoreEmpty`.  This defaults to true and tells cosmiconfig whether to continue searching
+for files when an empty one is found.
+
 #### `load([configPath])`
 
 Loads a configuration file directly.  Rejects with an error if the file does not exist, if the file contents are empty, or if the file cannot be parsed.  Upon success it returns a Promise that resolves with an object with two properties:
 - `config`: The loaded and parsed configuration.
 - `filepath`: The filepath where this configuration was found.
 
+If the file is empty then `config` will be set to undefined and an additional property `isEmpty` will be set to true
```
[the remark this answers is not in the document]
sudo-suhas: Looks like you missed this one.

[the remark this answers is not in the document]
olsonpm: whoops - I think I only got the one by the searchOptions.  Didn't realize there were two

src/loadRc.js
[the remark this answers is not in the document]
sudo-suhas: My main intention is to avoid code repetition and what you suggested looks good to me.

test/failed-directories.test.js
sudo-suhas: No, let's keep the other ones. Might be useful if we ever refactor that section again.

test/failed-directories.test.js
sudo-suhas: Yes, that would be great. Tests are already quite verbose and numerous.

src/createExplorer.js
```diff
@@ -51,16 +51,31 @@ module.exports = function createExplorer(options: {
   }
 
   function search(
-    searchPath: string
+    searchPath: string,
+    searchOptions: {
+      ignoreEmpty?: boolean,
+    }
   ): Promise<?cosmiconfig$Result> | ?cosmiconfig$Result {
+    const sanitizedSearchOptions: {
```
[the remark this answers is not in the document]
olsonpm: I actually have never used types in javascript - only recently had to deal with them when working in various code-bases.  Thanks for the explanation, and I definitely don't have preferences surrounding types.

By the way, I swear I clicked on the 'view more' in the github interface for viewing your comments - but it looks like I missed a few.  Give me 15 minutes to make sure I got everything.

sudo-suhas:
The following comments were not addressed:
- https://github.com/davidtheclark/cosmiconfig/pull/107#discussion_r173060710
- https://github.com/davidtheclark/cosmiconfig/pull/107#discussion_r173060941
- https://github.com/davidtheclark/cosmiconfig/pull/107#discussion_r173061820
- https://github.com/davidtheclark/cosmiconfig/pull/107#discussion_r173062200
- https://github.com/davidtheclark/cosmiconfig/pull/107#discussion_r173062208
- https://github.com/davidtheclark/cosmiconfig/pull/107#discussion_r173062453

Maybe you missed them because they were collapsed?

src/createExplorer.js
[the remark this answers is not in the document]
sudo-suhas: No problem at all @olsonpm 😄. Take all the time you need!

src/loadDefinedFile.js
```diff
@@ -16,7 +16,11 @@ module.exports = function loadDefinedFile(
 ): Promise<?cosmiconfig$Result> | ?cosmiconfig$Result {
   function parseContent(content: ?string): ?cosmiconfig$Result {
     if (!content) {
-      throw new Error(`Config file is empty! Filepath - "${filepath}".`);
+      return {
```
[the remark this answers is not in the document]
olsonpm: I find reading top to bottom is easier to skim than left to right, but again I only care about that stuff in my own code bases.

olsonpm:
Okay - the latest commit addresses

- [the repetitive parsing code](https://github.com/davidtheclark/cosmiconfig/pull/107#discussion_r173063293)
- [remove unnecessary yml test](https://github.com/davidtheclark/cosmiconfig/pull/107#discussion_r173065153)
- [add empty files in existing tests to ensure intended default 'ignoreEmpty' behavior](https://github.com/davidtheclark/cosmiconfig/pull/107#discussion_r173065304)

The commit prior to that should have addressed the other small changes that I missed per your earlier comment.  Again, sorry to cause the unnecessary back and forth.

src/loadRc.js
```diff
@@ -31,73 +41,32 @@ module.exports = function loadRc(
...
+      const fpath = `${filepath}.${extension}`;
+      const parseRcFile = createParseFile(fpath, parse, options.ignoreEmpty);
+
+      // Check the result from the previous `loadExtension` invocation. If result
+      // isn't null, just return that.
+      return result => {
+        if (result != null) return result;
+
+        // Try to load the rc file for the given extension.
+        return funcRunner(readRcFile(extension), [parseRcFile]);
```
sudo-suhas: Lets pass the `fpath` we constructed here instead of the extension which is used to build that again in `readRcFile`.

test/failed-directories.test.js
```diff
@@ -276,9 +344,10 @@ describe('cosmiconfig', () => {
           switch (searchPath) {
             case absolutePath('a/b/c/d/e/f/package.json'):
             case absolutePath('a/b/c/d/e/f/.foorc'):
+              throw { code: 'ENOENT' };
             case absolutePath('a/b/c/d/e/f/.foorc.json'):
             case absolutePath('a/b/c/d/e/f/.foorc.yaml'):
-              throw { code: 'ENOENT' };
+              return '';
```
sudo-suhas: Could you add a comment explaining what this tests(since it is multi-purpose)?

src/createParseFile.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+module.exports = function createParseFile(filepath, parse, ignoreEmpty) {
```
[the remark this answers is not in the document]
olsonpm: Sounds good.

As for the type annotations - I personally don't want to invest too much time learning any javascript typing.  I definitely don't mind glancing at the docs and making sure the code works, but if there are further modifications I'd appreciate either you or davidtheclark making the final adjustments.

olsonpm:
done

olsonpm:
not a problem.  I really appreciate your patience

sudo-suhas:
> the preference was regarding variable names. I've been going verbose lately

This is purely academic discussion at this point but I was in the same boat as you. However, I have come to grudgingly concede that short variable names are not bad as we have been told. And in fact, having short variable names makes a lot of sense in some situations. See https://talks.golang.org/2014/names.slide#1 if you are interested.

olsonpm:
haha.  Thanks for the link.  Who knows how I'll code two years from now but I don't see myself straying toward some of these opinions e.g.

> A good name is
> - easy to type

or
> Prefer r to reader. 
> Prefer b to buffer.

as I have a moral stance against single letter shell options.  I understand that one can learn to read these, but they are hardly intuitive.  One thing that drove me nuts earlier was learning why on earth vue decided `createElement` should be aliased as `h` [in their documentation](https://vuejs.org/v2/guide/render-function.html#JSX).  A couple angry googles led me [to the source](https://github.com/vuejs/babel-plugin-transform-vue-jsx/issues/6), only to still wonder "is typing `createElement` really so inconvenient that they need a single, seemingly unassociated letter to replace it?"

There's definitely a class of programmers who highly value conciseness and my guess is they tend to have better memories than us mortals.  I'm convinced they are the ones who created `sh` and `bash` and are to blame for all the [insane variable manipulations cryptically available](https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion)

Anyway, I'm currently focused on writing code that reads english'y.  I understand it's not possible in certain circumstances due to english hardly being computer friendly - and lower level operations or code focused on performance will never read well, but their overall operations can easily be wrapped in a nice intuitive function name where its internals are shielded from innocent eyes.

davidtheclark:
Thanks for all the work. I'm going to merge this into a v5 branch and we can follow up with other details, if we'd like before release.

@sudo-suhas I dunno ... in that slideshow you linked, the only reason I found for preferring short names was "long names obscure what the code does". And I don't understand that justification. 🤷‍♂️

davidtheclark:
@sudo-suhas @olsonpm I pushed a v5 branch and forced master back to the 4.0.0 tag. You'll probably need to delete your local master and re-pull it next time you need it.

sudo-suhas:
@davidtheclark I understand what you mean. I personally prefer neither extreme. In some cases, the short variable names conveys precise information about it's role(like `i` instead of `index` or `idx`). Furthermore, I think some of these conventions apply more for statically typed language where the variable declaration provides good amount of information about what the variable is for.

sudo-suhas:
> You'll probably need to delete your local master and re-pull it

`git pull --rebase` takes care of it just fine.

sudo-suhas:
@davidtheclark What are your thoughts on this - https://github.com/davidtheclark/cosmiconfig/pull/107#issuecomment-371377662

davidtheclark:
> @davidtheclark What are your thoughts on this - #107 (comment)

Thanks for pointing that out. 

Maybe we should remove the `format` option for `load`. The amount of time it might save (by avoiding an attempt at parsing JS as YAML) would be so inconsequential.

And for `searchPath[, searchOptions]`, I'm fine with the approach "in the javascript wild" of collapsing the arguments. But I think @olsonpm's suggestion to just have a single options argument for both functions is the nicest. If you're ok with that, I'll work that in.

sudo-suhas:
> Maybe we should remove the `format` option for `load`.

I am okay with this.

> But I think @olsonpm's suggestion to just have a single options argument for both functions is the nicest.

I too agree, let's do it

PR #108, opened 2018-02-09
WIP: Split load into two functions

olsonpm:
- Make 'load' take a single optional 'searchPath' parameter.  It
  searches up the directory tree until a configuration is found.

- Add 'loadConfig' which takes an optional 'configPath' parameter and
  attempts to parse it.

- Modify all tests accordingly and add a few resulting from the changes


olsonpm:
This addresses [the comment here](https://github.com/davidtheclark/cosmiconfig/issues/106#issuecomment-363811397)

I purposefully left out documentation changes until both wording and implementation was landed on.

davidtheclark:
@olsonpm: Thanks. I like it! @sudo-suhas what do you think?

test/successful-directories.test.js
```diff
@@ -300,15 +303,53 @@ describe('cosmiconfig', () => {
...
+          case absolutePath('a/b/c/d/e/.foorc'):
+            throw { code: 'ENOENT' };
+          case absolutePath('a/b/c/d/e/foo.config.js'):
+            return 'module.exports = { found: true };';
+          default:
+            throw new Error(`irrelevant path ${searchPath}`);
+        }
+      }
+      const readFileMock = mockReadFile(sync, readFile);
+      process.cwd = jest.fn(() => absolutePath('a/b/c/d/e/f'));
```
sudo-suhas: Do we need to revert this change in a `finally`?

sudo-suhas:
I too am in agreement for the proposed changes. Thank you @olsonpm for your work on this!

I don't anticipate any difficulty in terms of API change for libraries which depend on `cosmiconfig`. @davidtheclark Could you confirm this w.r.t `stylelint`? I'd also like to hear @azz's thoughts on this(_Right now, prettier [exposes and uses only the `load` function](https://github.com/prettier/prettier/blob/1ca4e2b90a2c514e094041700b24c633dd08624f/src/config/resolve-config.js#L28) from the explorer instance._).

test/successful-directories.test.js
olsonpm: I'm not super familiar with testing frameworks, but does [`afterEach`](https://github.com/davidtheclark/cosmiconfig/pull/108/files/5401cba51c0a1887b6a32e7012f86eb68e4e1e6f#diff-4b7ed0b193d73a1b1fa66c79c737b673R30) not cut it?

olsonpm:
I can run some integration tests with those libraries.  I've been working a little with `prettier` and the libraries it integrates with so nothing should be too new for me.  I'll edit this post with what I find.

test/successful-directories.test.js
sudo-suhas: My apologies, I did not see that. `afterEach` will suffice. Though I have some reservations about using the test hooks for this since we only mock `process.cwd` in one test.

olsonpm: makes sense, i'll wrap it in a try/catch

azz:
API change sounds fine to me, unless I'm misreading we won't have to change anything?

olsonpm:
if it does i'll create a prettier PR with the changes and it here

sudo-suhas:
@azz I think changes would be required to accommodate the [`--config` flag](https://prettier.io/docs/en/cli.html#find-config-path-and-config) because you would need to use `search` for the usual search and `load` for loading a specific config file by location.

Arcanemagus:
> API change sounds fine to me. Unless I'm misreading we won't have to change anything?
> if it does i'll create a prettier PR with the changes and link it here

Note that there are [_far_ more](https://www.npmjs.com/browse/depended/cosmiconfig) than just `stylelint` and `prettier` using this, so make sure any API changes follow semver when published 😉.

sudo-suhas:
@Arcanemagus these changes would be published with a major version change so no need to worry on that front. The intention behind discussing this with maintainers/authors of `stylelint` and `stylelint` is to ensure that we are not overlooking scenarios which might affect _any_ of the dependents. It would certainly help if developers watched the repo and provided appropriate feedback for PRs like this one and #107.

olsonpm:
I actually spent a good amount of time today trying to find ways to query packages that

- depend on cosmiconfig
- sort by downloads

to find major repos worth making sure the changes work.  Unfortunately getting that data didn't look easy.

sudo-suhas:
Thanks @olsonpm for all the awesome work you are putting into this! Appreciate it 👍

Arcanemagus:
@olsonpm Unfortunately it looks like you would have to load the dependent list and then query each individually for their download count. Seems like an oversight on the part of NPM here 😢.

davidtheclark:
This seems to me like a pretty straightforward clarification of the API. If we publish it as a major version upgrade, I'm not worried about users having trouble dealing with it. They'll simply need to determine which functionality they wanted and then change the invocation — behavior and results will be exactly the same.

So I don't think we need to worry about dependents here. They'll be fine. The only reason I think we might not want to make the change is if people don't think the clarification is worth the slight burden when upgrading. If everybody paying attention to this repo thinks the clarification is worthwhile, let's do it.

sudo-suhas:
Great. @olsonpm could you please update the documentation and the changelog?

olsonpm:
yup.  sunday night or monday i'll be on it

davidtheclark:
Awesome, @olsonpm. I think this is ready to merge, right? Can you and @sudo-suhas give me a 👍if so — want to make sure I'm not missing something.

README.md
```diff
@@ -173,54 +173,60 @@ A function that transforms the parsed configuration. Receives the result object
...
+- `filepath`: The filepath where this configuration was found.
+
+```js
+explorer.load() // loads options.configPath
+explorer.load('load/this/file.json');
+```
+
+Note the [`format` option](#format) is applicable for this.
 
 #### `clearFileCache()`
```
sudo-suhas: Shall we rename these to `clearLoadCache` and `clearSearchCache`?

sudo-suhas:
My sincere apologies! I had not realised that @olsonpm had updated the readme and the changelog. Apart from the minor renaming suggestion(_would be a breaking change, needs to be mentioned in changelog_), it looks great.

olsonpm:
the only intelligent word i have atm is "give me til monday.  Please"

README.md
davidtheclark: 👍

olsonpm: I guess they're synonymous i.e. the loadCache will only ever cache files and likewise with search <--> directories so renaming makes sense.  I'll push that change soon.

olsonpm:
Yeah sorry @sudo-suhas, I didn't ping after I pushed the update because I thought my pushed changes would ping everyone subscribed to this thread.

olsonpm:
I'm tidying that one up now and will definitely want a pair of eyes on the implementation in case I misinterpret what you guys are looking for.

PR #111, opened 2018-03-11
Improve documentation for v5

davidtheclark:
I took a pass through the README documentation.

I also added a remark-cli step that lints the README (with autocorrect when possible). I've found this especially nice to catch broken links.

@sudo-suhas or @olsonpm for review, please.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 
 - Removed: support for loading config path using the `--config` flag. cosmiconfig will not parse command line arguments. Your application can parse command line arguments and pass them to cosmiconfig.
 - Removed: `argv` config option.
-- Removed: support for Node versions < 4.
+- Removed: support for Node versions &lt; 4.
 - Added: `sync` option.
 - Fixed: Throw a clear error on getting empty config file.
 - Fixed: when a `options.configPath` is `package.json`, return the package prop, not the entire JSON file.
@@ -68,7 +68,7 @@
 
 - Changed: module now creates cosmiconfig instances with `load` methods (see README).
 - Added: caching (enabled by the change above).
-- Removed: support for Node versions <4.
+- Removed: support for Node versions &lt;4.
 
 ## 1.1.0
 
@@ -87,4 +87,5 @@
 - Initial release.
 
 [parse-json-pr-12]: https://github.com/sindresorhus/parse-json/pull/12
+
 [pr-101]: https://github.com/davidtheclark/cosmiconfig/pull/101
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ Default: `process.cwd()`.
 Type: `boolean`.
 Default: `true`.
 
-By default, `search` ignores empty configuration files and continues searching up the tree. If the `search` option [`ignoreEmpty`] is set to `false` and an empty configuration file is found, the [result] will include `config: undefined` and `isEmpty: true`.
+By default, `search` ignores empty configuration files and continues searching up the tree. If this option is `false` and an empty configuration file is found, the [result] will include `config: undefined` and `isEmpty: true`.
 
 ### explorer.load
 
--- a/README.md
+++ b/README.md
@@ -3,11 +3,13 @@
 [![Build Status](https://img.shields.io/travis/davidtheclark/cosmiconfig/master.svg?label=unix%20build)](https://travis-ci.org/davidtheclark/cosmiconfig) [![Build status](https://img.shields.io/appveyor/ci/davidtheclark/cosmiconfig/master.svg?label=windows%20build)](https://ci.appveyor.com/project/davidtheclark/cosmiconfig/branch/master)
 
 Find and load a configuration object from
+
 - a `package.json` property (anywhere up the directory tree)
 - a JSON or YAML "rc file" (anywhere up the directory tree)
 - a `.config.js` CommonJS module (anywhere up the directory tree)
 
 For example, if your module's name is "soursocks," cosmiconfig will search out configuration in the following places:
+
 - a `soursocks` property in `package.json` (anywhere up the directory tree)
 - a `.soursocksrc` file in JSON or YAML format (anywhere up the directory tree)
 - a `soursocks.config.js` file exporting a JS object (anywhere up the directory tree)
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -29,99 +31,99 @@ Tested in Node 4+.
...
-Default: `'.[moduleName]rc'`
+Type: `string` or `false`.
+Default: `\`.${moduleName}rc\`\`.
 
 Name of the "rc file" to look for, which can be formatted as JSON or YAML.
 
 If `false`, cosmiconfig will not look for an rc file.
 
-If `rcExtensions: true`, the rc file can also have extensions that specify the syntax, e.g. `.[moduleName]rc.json`.
+If `rcExtensions: true`, the rc file can also have extensions that specify the syntax, e.g. `.${moduleName]rc.json`.
```
olsonpm:
```txt
${moduleName]rc.json
            ^
```

README.md
```diff
@@ -145,104 +147,129 @@ Instead of *just* looking for `.goldengrahamsrc` (no extension), it will also lo
...
+##### ignoreEmpty
 
-#### `load([configPath])`
+Type: `boolean`.
+Default: `true`.
 
-Loads a configuration file directly.  Rejects with an error if the file does not exist, if the file contents are empty, or if the file cannot be parsed.  Upon success it returns a Promise that resolves with an object with two properties:
-- `config`: The loaded and parsed configuration.
-- `filepath`: The filepath where this configuration was found.
+By default, `search`, ignores empty configuration files and continues searching up the tree. If the `search` option [`ignoreEmpty`] is set to `false` and an empty configuration file is found, the [result] will have include `config: undefined` and `isEmpty: true`.
```
olsonpm: I'm the worst with commas, but just checking to see if you intended on the comma after `search`

```txt
By default, `search`, ignores
                    ^
```

README.md
```diff
@@ -29,99 +31,99 @@ Tested in Node 4+.
...
-## API
+### Create an explorer
 
-### `var explorer = cosmiconfig(moduleName[, options])`
+```js
+const explorer = cosmiconfig(moduleName[, cosmiconfigOptions])
+```
 
-Creates a cosmiconfig instance (i.e. explorer) configured according to the arguments, and initializes its caches.
+Creates a cosmiconfig instance ("explorer") configured according to the arguments, and initializes its caches.
```
olsonpm: Not sure whether you and @sudosuhas already discussed this, but do you have thoughts on normalizing the terminology surrounding **cosmiconfig instance** and **explorer**?  Another term used right under '## Usage' is **cosmiconfig explorer**.

This might be solved by an api change with the method 'createExplorer' being a property of cosmiconfig.  Anything besides that leaves room for ambiguity - which is not a big deal.  Just wanted to see what you guys thought.

davidtheclark: I don't think we need to make any API change; but if we wanted to change the docs we could do this:

```js
// Change this:
const cosmiconfig = require('cosmiconfig');
// To this:
const createExplorer = require('cosmiconfig');
```

olsonpm: haha true

README.md
```diff
@@ -29,99 +31,99 @@ Tested in Node 4+.
...
+
+// Search for a configuration by walking up directories.
+// See documentation for search, below.
+explorer.search().then(
+  (result) => {
+    // result.config is the parsed configuration object.
+    // result.filepath is the path to the config file that was found.
+    // result.isEmpty is true if there was nothing to parse in the config file.
+  },
+  (parsingError) => {
```
sudo-suhas: I prefer the `.catch` equivalent as it seems clearer and explicit(to me at least).

README.md
```diff
@@ -29,99 +31,99 @@ Tested in Node 4+.
...
 
-#### Options
+#### cosmiconfigOptions
 
 ##### packageProp
 
-Type: `string` or `false`
-Default: `'[moduleName]'`
+Type: `string` or `false`.
+Default: `'${moduleName}'`.
```
sudo-suhas: This still renders to a single line. If we want the `Type` and `Default` to be on different lines, we should either use `<br>` or bulletting.

README.md
```diff
@@ -29,99 +31,99 @@ Tested in Node 4+.
...
 Name of the property in `package.json` to look for.
 
 If `false`, cosmiconfig will not look in `package.json` files.
 
 ##### rc
 
-Type: `string` or `false`
-Default: `'.[moduleName]rc'`
+Type: `string` or `false`.
+Default: `\`.${moduleName}rc\`\`.
```
sudo-suhas: The escaping does not work. This is how it renders:

image

README.md
```diff
@@ -29,99 +31,99 @@ Tested in Node 4+.
...
 You may like extensions on your rc files because you'll get syntax highlighting and linting in text editors.
-Also, with `rcExtensions: true`, you can use JS modules as rc files, e.g. `.[moduleName]rc.js`.
+Also, with `rcExtensions: true`, you can use JS modules as rc files, e.g. `.${moduleName}rc.js`.
 
 ##### js
 
-Type: `string` or `false`
-Default: `'[moduleName].config.js'`
+Type: `string` or `false`.
+Default: `\`${moduleName}.config.js\`\`.
```
sudo-suhas: Escaping needs to be removed here too.

README.md
```diff
@@ -145,104 +147,129 @@ Instead of *just* looking for `.goldengrahamsrc` (no extension), it will also lo
...
+explorer.search([searchPath][, searchOptions])
+```
 
-Searches for a configuration file. Returns a Promise that resolves with `null`, if nothing is found, or an object with two properties:
-- `config`: The loaded and parsed configuration.
-- `filepath`: The filepath where this configuration was found.
+Searches for a configuration file. In async mode (default), returns a Promise that resolves a [result] object or with `null`, if no configuration file is found. In [`sync`] mode, returns a [result] object or `null`.
 
-When the [search option](#searchoptions) `ignoreEmpty` is set to false and an empty configuration file is found, the result will have `config` set to `undefined` and an additional property `isEmpty` set to true.
+So let's say `const explorer = cosmiconfig('goldengrahams');` — here's how `explorer.search` will work:
```
sudo-suhas: Change to "here's how `explorer.search` works" to be consistent with the rest of the explanation.

README.md
```diff
@@ -145,104 +147,129 @@ Instead of *just* looking for `.goldengrahamsrc` (no extension), it will also lo
...
+##### ignoreEmpty
 
-#### `load([configPath])`
+Type: `boolean`.
+Default: `true`.
 
-Loads a configuration file directly.  Rejects with an error if the file does not exist, if the file contents are empty, or if the file cannot be parsed.  Upon success it returns a Promise that resolves with an object with two properties:
-- `config`: The loaded and parsed configuration.
-- `filepath`: The filepath where this configuration was found.
+By default, `search` ignores empty configuration files and continues searching up the tree. If the `search` option [`ignoreEmpty`] is set to `false` and an empty configuration file is found, the [result] will have include `config: undefined` and `isEmpty: true`.
```
sudo-suhas: `"will have include" => "will have"`

README.md
```diff
@@ -145,104 +147,129 @@ Instead of *just* looking for `.goldengrahamsrc` (no extension), it will also lo
...
+
+Clears the cache used in [`search()`].
+
+### explorer.clearCaches
+
+Performs both [`clearLoadCache()`] and [`clearSearchCache()`].
+
+## Caching
+
+As of v2, cosmiconfig uses a few caches to reduce the need for repetitious reading of the filesystem. Every new cosmiconfig instance (created with `cosmiconfig()`) has its own caches.
```
sudo-suhas: `"uses a few caches" => "uses caching"`

README.md
```diff
@@ -29,99 +31,99 @@ Tested in Node 4+.
...
-As of v2, cosmiconfig uses a few caches to reduce the need for repetitious reading of the filesystem. Every new cosmiconfig instance (created with `cosmiconfig()`) has its own caches.
+The result object you get from `search` or `load` has the following properties:
 
-To avoid or work around caching, you can
-- create separate instances of cosmiconfig, or
-- set `cache: false` in your options.
-- use the cache clearing methods documented below.
+- **config:** The parsed configuration object. `undefined` if the file is empty.
+- **filepath:** The path to the configuration file that was found.
+- **isEmpty:** `true` if the configuration file is empty.
```
sudo-suhas: Should we mention that this property will not be present if the search was successful?

sudo-suhas:
What do you think of adding a "Used By" section to the readme?

CHANGELOG.md
```diff
@@ -26,7 +26,7 @@
 
 - Removed: support for loading config path using the `--config` flag. cosmiconfig will not parse command line arguments. Your application can parse command line arguments and pass them to cosmiconfig.
 - Removed: `argv` config option.
-- Removed: support for Node versions < 4.
+- Removed: support for Node versions &lt; 4.
```
sudo-suhas: Is it possible to use the `<` symbol instead of `&lt;`? Seems unnecessary..

davidtheclark: This is automatically changed by remark-cli, so I'm not going to worry about it.

README.md
davidtheclark: It's less precise, though: if your callback that does something with `result` throws an error, that error is going to end up in the `catch` statement but will not be a parsing error.

README.md
davidtheclark: Yeah, I think that's fine. I often type Markdown with individual sentences on separate lines to make the diff more meaningful.

README.md
davidtheclark: Bummer.

README.md
davidtheclark: Sure.

README.md
davidtheclark: I'll change it to `catch` and just change the `parsingError` name to `error`.

davidtheclark:
@sudo-suhas Thanks for catching those typos and for the suggestions. I addressed the feedback in the new commit.

If you'd like to add a "Used By" section to the README, sounds good to me 👍

README.md
```diff
@@ -145,104 +146,129 @@ Instead of *just* looking for `.goldengrahamsrc` (no extension), it will also lo
...
-##### searchOptions
+##### ignoreEmpty
 
-Can be an object containing a single property `ignoreEmpty`.  This defaults to true and tells cosmiconfig whether to continue searching
-for files when an empty one is found.
+Type: `boolean`.
+Default: `true`.
 
-#### `load([configPath])`
+By default, `search` ignores empty configuration files and continues searching up the tree. If the `search` option [`ignoreEmpty`] is set to `false` and an empty configuration file is found, the [result] will include `config: undefined` and `isEmpty: true`.
```
sudo-suhas: This links back to itself. Is that necessary?

Issue #109, opened 2018-02-14
Load an ES2015 module config file 


danawoodman:
I am writing a CLI tool where I want my end users to be able to write their config file in ES6/7/Next, how can I accomplish this?

Eg, my `foo.config.js` file:
```js
import foo from './foo'

export default {  foo, someOtherOption: true }
```

danawoodman:
I just implemented this in one of my own CLI tools and it should be as easy as importing `babel-register` and then importing the `default` export of its and ES module

danawoodman:
Had to stop using cosmiconfig because my tool needs to support ES projects which may define variables in files using ES modules which are then included into the config file.

davidtheclark:
Open to ideas here!

danawoodman:
Basically I did this:

```js
import 'babel-register'

function importModule(module) {
  return module.__esModule ? module.default : module
}

const config = importModule(require(path.join(process.cwd(), 'myapp.config.js')))
```

davidtheclark:
I changed the title of this issue to reflect that it is a new feature request. PRs welcome.

chrisblossom:
This is how I support ``es modules`` with ``cosmiconfig`` (I think got the idea from eslint or babel). Maybe this should be handled by the end user because otherwise ``export default`` is only supported. If that is okay/expected, I could put in a PR to address this issue. Would require a major version bump because it would be a breaking change though.

```js
import cosmiconfig from 'cosmiconfig';

function transform(result) {
    let config = result.config;

    /**
     * Handle ES Modules
     */
    if (typeof config === 'object' && config.__esModule) {
        if (config.default) {
            config = config.default;
        } else {
            throw new Error(
                `${result.filepath} must use default export with ES Modules`,
            );
        }
    }

    return {
        ...result,
        config,
    };
}

const explorer = cosmiconfig('cfg', {
    rcExtensions: true,
    sync: true,
    transform,
});

function loadOptionsFile() {
    const { config } = explorer.load();

    return config;
}
```

danawoodman:
What I mentioned above should be simpler and not require a major version bump AFAIK

chrisblossom:
@danawoodman we solved it the same way, I just gave a full example to use with cosmiconfig's transform option. The reason both require a version bump is because if someone is using a non-default export it would break their code. If this is expected, then a major version bump isn't required.

danawoodman:
Why would someone use a non-default export if using ES modules?

[...]

PR #113, opened 2018-03-12
add support for precompiled modules

chrisblossom:
PR for https://github.com/davidtheclark/cosmiconfig/issues/109.

I think this is missing some tests for files ``test/successful-directories.test.js`` ``test/failed-directories.test.js``. I am not sure how those tests work.

If there is anything else missing, or you'd like something changed just let me know! It currently throws when es modules are detected without a default export.

If you'd rather this be PR to master, I have that ready as well.

### DIFF
--- a/src/loadDefinedFile.js
+++ b/src/loadDefinedFile.js
@@ -2,7 +2,7 @@
 'use strict';
 
 const yaml = require('js-yaml');
-const requireFromString = require('require-from-string');
+const requireJs = require('./requireJs');
 const readFile = require('./readFile');
 const parseJson = require('./parseJson');
 const path = require('path');
@@ -31,7 +31,7 @@ module.exports = function loadDefinedFile(
         });
         break;
       case 'js':
-        parsedConfig = requireFromString(content, filepath);
+        parsedConfig = requireJs(content, filepath);
         break;
       default:
         parsedConfig = tryAllParsing(content, filepath);
@@ -91,8 +91,12 @@ function tryYaml(content: string, filepath: string, cb: () => ?Object) {
 
 function tryRequire(content: string, filepath: string, cb: () => ?Object) {
   try {
-    return requireFromString(content, filepath);
+    return requireJs(content, filepath);
   } catch (e) {
+    if (e.code === 'ES_MODULES_DEFAULT_REQUIRED') {
+      throw e;
+    }
+
     return cb();
   }
 }
--- a/src/requireJs.js
+++ b/src/requireJs.js
@@ -1,0 +1,30 @@
+// @flow
+'use strict';
+
+const requireFromString = require('require-from-string');
+
+function requireJs(content: string, filepath: string) {
+  let result = requireFromString(content, filepath);
+
+  /**
+   * Handle ES Modules
+   */
+  if (typeof result === 'object' && result.__esModule) {
+    if (result.default) {
+      result = result.default;
+    } else {
+      const error = new Error(
+        `${filepath} must use default export with ES Modules`
+      );
+
+      // $FlowFixMe
+      error.code = 'ES_MODULES_DEFAULT_REQUIRED';
+
+      throw error;
+    }
+  }
+
+  return result;
+}
+
+module.exports = requireJs;
--- a/test/failed-files.test.js
+++ b/test/failed-files.test.js
@@ -120,6 +120,8 @@ describe('cosmiconfig', () => {
         makeEmptyFileTest('js')
       );
 
+      it('throws error for empty file, format JS', makeEmptyFileTest('js'));
+
       it(
         'returns an empty config result for an empty file, format JSON',
         makeEmptyFileTest('json')
@@ -178,6 +180,11 @@ describe('cosmiconfig', () => {
         'throws error if defined JS file has syntax error',
         makeSyntaxErrWithoutKnownExtnameTest('js')
       );
+
+      it(
+        'throws error if defined JS ES Modules file has syntax error',
+        makeSyntaxErrWithoutKnownExtnameTest('cjs-es-module-js')
+      );
     });
 
     it('returns null if configuration file does not exist', () => {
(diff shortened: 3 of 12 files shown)


olsonpm:
So I don't think this addresses #109 since that issue wants support for direct (not precompiled) es modules.  This is still useful as it makes sense to add support for precompiled modules.

test/failed-files.test.js
```diff
@@ -16,6 +16,9 @@ const expectForFormat = {
   js(err) {
     expect(err.name).toBe('ReferenceError');
   },
+  'es-module'(err) {
```
olsonpm: so `format` in this code is referring to [these](https://github.com/davidtheclark/cosmiconfig/blob/master/src/createExplorer.js#L16).  The purpose of format is to tell cosmiconfig not to [guess and check](https://github.com/davidtheclark/cosmiconfig/blob/master/src/loadDefinedFile.js#L37) the different formats.

You were right not to add `es-module` as a format and so I think this test should be done a different way.  I think a one-off test [underneath this describe block](https://github.com/davidtheclark/cosmiconfig/blob/v5/test/failed-files.test.js#L88) with the same logic you created should do.

test/successful-directories.test.js
```diff
@@ -156,6 +156,52 @@ describe('cosmiconfig', () => {
...
+          switch (searchPath) {
+            case absolutePath('a/b/c/d/e/f/package.json'):
+            case absolutePath('a/b/c/d/e/f/.foorc'):
+            case absolutePath('a/b/c/d/e/package.json'):
+            case absolutePath('a/b/c/d/e/.foorc'):
+            case absolutePath('a/b/c/d/e/foo.config.js'):
+              throw { code: 'ENOENT' };
+            case absolutePath('a/b/c/d/e/f/foo.config.js'):
+              return `
+Object.defineProperty(exports, '__esModule', {
```
olsonpm: I *would* suggest shoving that text down to the bottom in a function e.g.

```js
getFooConfigJs() {
  return `
...
`
}
```

but there's probably a better way - so I *do* suggest to hang tight til davidtheclark or sudo-suhas gets around to recommending something better ha.

olsonpm:
What you modified is exactly what I had in mind.  Thanks for the PR and the changes.

chrisblossom:
I agree this does not address #109. I'd like to build off this PR and thought it would be too many features in one PR otherwise.

src/loadDefinedFile.js
```diff
@@ -91,8 +91,12 @@ function tryYaml(content: string, filepath: string, cb: () => ?Object) {
 
 function tryRequire(content: string, filepath: string, cb: () => ?Object) {
   try {
-    return requireFromString(content, filepath);
+    return requireJs(content, filepath);
   } catch (e) {
+    if (e.message.indexOf('must use default export with ES Modules') !== -1) {
+      throw e;
+    }
+
```
chrisblossom: @olsonpm I think this is expected/wanted behavior here. "[throws an error if non-default es module with missing extension](https://github.com/davidtheclark/cosmiconfig/pull/113/files#diff-06c2c9f7946f98248f0f685ee7c0960eR245)" is the respective test. Otherwise "Failed to parse" error is thrown. Please confirm.

test/successful-directories.test.js
chrisblossom: I am not sure I understand what you are suggesting here. I copied the format from https://github.com/davidtheclark/cosmiconfig/blob/44a65ab84dd223b231e55e6e3752b8736564d872/test/successful-directories.test.js#L134

chrisblossom:
Also, let me know if I should rebase the commits into one before this is merged.

src/loadDefinedFile.js
olsonpm: that would be my expectation.  However support for "precompiled modules" as a whole may not be what sudo-suhas and davidtheclark have in mind - so it'd be best to wait for them before figuring out specifics.

test/successful-directories.test.js
olsonpm: yeah that makes sense.  For some reason the 8 lines of template text seems to get in the way for me - but that's just preference and mine will probably conflict with what the other guys think.  No worries

src/loadDefinedFile.js
chrisblossom: https://github.com/davidtheclark/cosmiconfig/issues/109#issuecomment-372137187 leaves me with the impression that is what is wanted. I definitely could be misunderstanding though.

Thanks for the help with this PR! I appreciate it.

olsonpm: Ah - yeah my interpretation was not to worry about looking for named exports for a configuration.  Your interpretation makes more sense though :)

test/successful-directories.test.js
chrisblossom: I made it two lines. Better?

src/requireJs.js
```diff
@@ -0,0 +1,30 @@
...
+  if (typeof result === 'object' && result.__esModule) {
+    if (result.default) {
+      result = result.default;
+    } else {
+      const error = new Error(
+        `${filepath} must use default export with ES Modules`
+      );
+
+      // $FlowFixMe
+      error.code = 'ES_MODULES_DEFAULT_REQUIRED';
```
chrisblossom: @sudo-suhas I've added an error code here. Flow doesn't like it when you extend ``Error``. Is the ``// $FlowFixMe`` acceptable or should there be a different solution?

sudo-suhas: Setting the code on the instance looks fine to me. I don't think [extending the error class is recommended either](https://github.com/i0natan/nodebestpractices#-22-use-only-the-built-in-error-object). I am not sure if `$FlowFixMe` is required(haven't used flow much myself). @davidtheclark should be able to answer it better.

chrisblossom: > Can we set and check for a code prop on the error?

@sudo-suhas I added the ``error.code`` here at your request. Did I miss understand what you were asking, or am I incorrectly adding an error code?

sudo-suhas: No, no misunderstanding, you are correctly setting the error code.

davidtheclark:
@chrisblossom Is this PR obsolete because of the decision in https://github.com/davidtheclark/cosmiconfig/issues/109 to try exposing a `loadJs` function?

sudo-suhas:
I think some example code or a demo pkg should be sufficient to show how you should use ES modules with cosmiconfig. So I would say this PR is not necessary at least not in it's current form.

chrisblossom:
100% agree that this PR should be closed in the current form it is in. Closing for now and we can revisit this later.

PR #114, opened 2018-03-13
docs: fix typo

poppinlp:
Little typo.


davidtheclark:
Thanks @poppinlp. Whether it's a "typo" or not is debatable, since comma inside quotation marks is the American standard — but I'm happy to switch 😄

poppinlp:
ahha... Sry for that i'm not native speaker. But learned something new. Interesting :)

PR #116, opened 2018-03-13
add wallaby.js configuration file

chrisblossom:
I don't expect this to be merged, but I use this to run tests directly inside my IDE. Thought it might be helpful for others. https://wallabyjs.com/

Close if not interested/wanted.

### DIFF
--- a/wallaby.config.js
+++ b/wallaby.config.js
@@ -1,0 +1,44 @@
+'use strict';
+
+module.exports = wallaby => {
+  return {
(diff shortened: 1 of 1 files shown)


davidtheclark:
Thanks for sharing @chrisblossom. I think I'll leave this out of the repo, so it's not part of the maintenance scope. But hopefully others who want to use wallaby.js can find this PR and copy your work 👍

Issue #112, opened 2018-03-11
Provide searchSync and loadSync functions, instead of the sync option


davidtheclark:
I was [reviewing the documentation](https://github.com/davidtheclark/cosmiconfig/pull/111), then poking around in the code a little bit, wondering if things could be simplified —— and I'm thinking that the challenges we have in documenting and Flow-typing functions, because they can be sync or async based on an option, indicates a flaw that we should refactor out.

Instead of the `sync` option, we could have `explorer.searchSync` and `explorer.loadSync` functions. Internally, we'd do the same thing with our utility functions. I believe that the standard practice across Node is to provide separate functions for sync and async versions of the same behavior — instead of overloading a single function. This results in each function having a cleaner signature, which is easier to document, type check, unit test.

Does anybody object? If not, I will work on this refactoring and the new functions before our v5 release.

olsonpm:
yeah that would follow node's approach (and thus others, since most libs I work with have a `*Sync` alternative.  big :+1:

olsonpm:
fyi - I think this implies a new option `transformSync`.

olsonpm:
yeah I thought a little about a clean approach - came up empty.  What you suggest sounds good to me as I also don't like the idea of two separate transform options.

sudo-suhas:
I too like the idea of having separate functions for the sync and async version. But should it be `searchSync` or `search.sync`? I have seen this pattern in various places including [`is-directory`](https://github.com/jonschlinkert/is-directory).

davidtheclark:
> But should it be searchSync or search.sync?

I believe that the reason libraries go with `fn` (async) and `fn.sync` is because of how CommonJS works: that way the module can have one main export, the default function that people should use most of the time, instead of exporting an object. So you get `fn` and `fn.sync` instead of `SomeModule.fn` and `SomeModule.fnSync`.

That situation doesn't apply to us, so I favor the `searchSync` convention.

sudo-suhas:
That makes sense @davidtheclark. Thanks for explaining.

davidtheclark:
#120 introduced the functions, but we still need a pass at documentation.

davidtheclark:
Documented now in the v5 branch.

PR #117, opened 2018-03-15
Test refactor (without source code change) in preparation for async-sync source code refactor

davidtheclark:
Some of the structures in the current tests are coupled to the `sync` option. We'll need to revise that in preparation for separate async-sync functions ( https://github.com/davidtheclark/cosmiconfig/issues/112). Also, I'd like to experiment with some refactoring (with the goal of simplifying code). But I don't want to do either of those things without ensuring that we can change the source code without significantly changing the tests. So this PR will be to refactor the tests *in preparation for code changes* (separate sync functions instead of a sync option, and refactoring), but without changing the source code at all.

Then in a follow-up PR, we can change the source code and only slightly adjust the tests. A pattern like this, in the adjusted tests represented by this first commit:

```js
  test('async', () => {
    const readFileMock = mockReadFile(false, readFile);
    return cosmiconfig('foo', { stopDir: absolutePath('.') })
      .load(startDir)
      .then(result => {
        checkResult(readFileMock, result);
      });
  });

  test('sync', () => {
    const readFileMock = mockReadFile(true, readFile);
    const result = cosmiconfig('foo', {
      stopDir: absolutePath('.'),
      sync: true,
    }).load(startDir);
    checkResult(readFileMock, result);
  });
```

Could then be slightly tweaked to look like this, when we revise the function signatures:

```js
  const explorer = cosmiconfig('foo', { stopDir: absolutePath('.') });

  test('async', () => {
    const readFileMock = mockReadFile(false, readFile);
    return explorer
      .load(startDir)
      .then(result => {
        checkResult(readFileMock, result);
      });
  });

  test('sync', () => {
    const readFileMock = mockReadFile(true, readFile);
    const result = explorer.loadSync(startDir);
    checkResult(readFileMock, result);
  });
```

I started with just one test file. Feedback welcome. I'll then move on to the others.

### DIFF
--- a/test/successful-directories.test.js
+++ b/test/successful-directories.test.js
@@ -9,8 +9,6 @@ const cosmiconfig = require('../src');
 
 const absolutePath = util.absolutePath;
 const mockReadFile = util.mockReadFile;
-const testFuncsRunner = util.testFuncsRunner;
-const testSyncAndAsync = util.testSyncAndAsync;
 
 beforeAll(() => {
   util.mockStatIsDirectory(true);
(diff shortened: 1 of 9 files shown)


olsonpm:
looks reasonable to me :+1:

chrisblossom:
As someone who just went through trying to understand how the tests work, this pattern is a lot easier to understand what is going on / what is being tested.

How do you feel about [removing mocks in place of spies to real file tests](https://github.com/chrisblossom/cosmiconfig/commit/06d546599a6d798f40dc89283790acb216e2c308)? Please read the comments on the bottom for reasoning/motivation. I think this would fit in nicely with this test refactor PR and I would be willing to rebase the changes to this PR if wanted.

test/successful-directories.test.js
```diff
@@ -31,447 +29,553 @@ afterAll(() => {
...
+    ]);
+
+    expect(result).toEqual({
+      config: { found: true },
+      filepath: absolutePath('a/b/c/d/.foorc'),
     });
+  };
+
+  test('async', () => {
+    const readFileMock = mockReadFile(false, readFile);
```
chrisblossom: Is there a point to instead just calling ``jest.spyOn`` directly here? Using ``jest.spyOn`` directly should simplify this.

```js
const readFileMock = jest.spyOn(fsMock, 'readFile').mockImplementation(readFile)
```

davidtheclark: Maybe a relic of pre-Jest days. I'd be fine with making this less indirect by using `jest.spyOn` directory 👍

davidtheclark:
@chrisblossom Since I'm thinking switching to regular `require` is the right move, I'd be fine with doing something like you suggest for using the real file system instead of mocking it. I'll need to finish this PR first (might be a few days, because I have family coming to visit today); then update it against v5 (where `load` and `search` are separated); then let's try the read-file-system approach. Sound good?

davidtheclark:
Ok, this branch is rebased against v5 and should incorporate the test changes from @olsonpm's v5 commits (separating `search` and `load`, normalizing the result with empty configs and adding the `ignoreEmpty` option).

I think other adjustments (like changing mocks or using the real file system) should be made in subsequent PRs instead of adding more to this one, which is already massive.

(Because this branch was originally based off master, it also brings a couple of small recent changes from master into v5. Very small, so I thought it wasn't worth Git-separating those out.)

If these tests look good to everyone (and pass on Appveyor), we can merge and then move on with subsequent adjustments. 

(I'll be mostly unavailable for a few days starting about now, so apologies for delays.)

chrisblossom:
@davidtheclark Sounds great!

test/successful-directories.test.js
sudo-suhas: The helper function allowed to use the same `readFile` implementation for both sync and async version(which involves callbacks).

sudo-suhas:
I understand that the way tests were written were hard to understand and follow(my bad). We had discussed the same earlier and ,I for the most part, I agree with the changes made here. However, I think some of the [tests are very similar](https://github.com/davidtheclark/cosmiconfig/blob/e39e3abcefddfe88b4ad14b1c8a88b870b20c863/test/successful-files.test.js#L8-L63) and could be re-written using  
[`jest-in-case`](https://github.com/atlassian/jest-in-case). I tend to take the de-duplication thing a bit too far so I understand if you do not want to incorporate this feedback.

chrisblossom:
@sudo-suhas personally I think that adds another layer of complexity that is unnecessary. Sure, it is less code, but makes it more difficult to understand. Usually you can load up a project's test and as long as you are familiar with the testing framework, you can understand how tests are being ran and where. When going through ``cosimiconfig``'s tests, I had to track down random test runners and try to understand what is going on. This PR does help a lot with that. Just my opinion / my experience though.

olsonpm:
I'm fine with whatever.  As I've said before, I tend to go with code deduplication by wrapping the repetitive stuff in functions with the goal of naming functions to make it easy to understand what's going on.  To me the "runAsyncAndSync" made it clear that the function was going to wrap the sync/async behavior - the content of the tests wouldn't change and were where I expected them to be.  Personally I didn't have a problem reading the tests.  I'm totally fine with explicit, duplicate blocks as well too.

sudo-suhas:
> personally I think that adds another layer of complexity that is unnecessary. Sure, it is less code, but makes it more difficult to understand. Usually you can load up a project's test and as long as you are familiar with the testing framework, you can understand how tests are being ran and where. When going through cosimiconfig's tests, I had to track down random test runners and try to understand what is going on.

@chrisblossom I concede on this point. However, I was trying to strike the optimum middle ground. But when I tried to change the test, I got errors because `jest` does not allow nested tests:
```
  ● loads defined config file › JSON

    Tests cannot be nested. Test `async` cannot run because it is nested within `JSON`.

      at Env.it (node_modules/jest-jasmine2/build/jasmine/Env.js:428:15)
      at Object.<anonymous>.opts (test/successful-files.test.js:17:3)
```

davidtheclark:
@sudo-suhas Thanks for the suggestions. I would like to go with this right now, favoring repetition, which has the benefit of test independence, over abstraction, which has the benefit of deduplication. I think the tests in this format are pretty easy to independently read, understand, amend, mimic, etc., so I don't think we yet have a *problem* that need to solve with abstraction and deduplication.

davidtheclark:
@sudo-suhas It looks like Travis and Appveyor are ignoring this PR. I imagine the reason is https://github.com/davidtheclark/cosmiconfig/pull/88. My inclination is to revert that commit, to ensure that tests are run against every PR, even if those PRs are not merging into master. Do you think there's a better way?

chrisblossom:
@davidtheclark I have sometime today to work on the real file tests / move to require / ``loadJs`` option. Is this PR at a point where that is safe to do / still wanted?

davidtheclark:
@chrisblossom Yes, I was just going to merge this but [realized the CI tests weren't running](https://github.com/davidtheclark/cosmiconfig/pull/117#issuecomment-375909706). So it might be failing on Appveyor 🤷‍♂️ , I don't know. Once @sudo-suhas makes a recommendation for how to handle that, I'll merge. So you could work off this branch until then?

sudo-suhas:
I added `v5` to the list of white listed branches to build on CI. This enables CI builds for PRs to the v5 branch. I also rebased this branch against the v5 branch to pull in the latest changes and run CI.

sudo-suhas:
CI build is failing because we are missing a test for this [/src/createExplorer.js@`2c5c9ce`#L141](https://github.com/davidtheclark/cosmiconfig/blob/2c5c9ce4e72cea1fac7cd242b502909698088624/src/createExplorer.js#L141).

PR #118, opened 2018-03-19
Remove and disable package-lock.json

sudo-suhas:
`package-lock.json` was added in #111 even though we have and use `yarn.lock`. This removes the `package-lock.json` file and adds a `.npmrc` file which disables creation of the lock file. npm 5 does not work on windows and the presence of it's lock file only causes more issues in CI in my experience. Which is why I'd vote for continuing to use yarn. This also updates the `yarn.lock` file for newly added deps.


olsonpm:
i think it's considered bad practice for a library/reusable module to hold a package-lock.json anyway.  I didn't look too deeply into why, but i'm pretty sure it helps with package duplication and reduces bloat. sindresorhus and shellscape both follow that pattern.  My understanding is that it's recommended for applications.

didn't know npm 5 had issues on windows.  Thanks for mentioning hat

davidtheclark:
Oops, my mistake.

I'm fine removing it and leaving `yarn.lock`. Just to clarify a couple points, though:

@sudo-suhas "npm 5 does not work on windows": I think a lot of people use it successfully on Windows; and it looks like the CI failures are about test coverage not installation failures, right? (e.g. https://ci.appveyor.com/project/davidtheclark/cosmiconfig/build/job/3erl53ej7wt4e5a1)

> i think it's considered bad practice for a library/reusable module to hold a package-lock.json anyway

It's encouraged to commit the lockfile to your repo (both [for Yarn](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/) and [for npm](https://docs.npmjs.com/files/package-lock.json)). Some people don't commit lockfiles in order to avoid the annoying merge conflicts they can produce when you change dependencies. But I think both Yarn and npm have since released ways to easily manage those conflicts. If a module like cosmiconfig is installed by an end user, its lockfile doesn't have any impact on the installation — just the semver ranges in `package.json`. I guess I'm not aware of outstanding reasons *not* to commit the lockfile.

olsonpm:
hmm - I can't find any good reasons to forego lock files then.  Sorry about that.  Maybe sindresorhus [just considers it noise](https://github.com/sindresorhus/delay/pull/23#issuecomment-350458370) like you mentioned

chrisblossom:
I am unsure about ``npm``, but ``yarn``  according to [yarnpkg.com/blog/2016/11/24/lockfiles-for-all](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/):  _"Lockfiles within your dependencies will be ignored"_.

I think the argument against using lockfiles in libraries would be if an upstream package breaks something, the library's tests wouldn't catch it, but the end user's app would break. Pretty much the opposite reason you want to use lock files in your application.

sudo-suhas:
> @sudo-suhas "npm 5 does not work on windows": I think a lot of people use it successfully on Windows; and it looks like the CI failures are about test coverage not installation failures, right?

We had faced issues due to npm in `lint-staged` (mostly when the `node_modules` cache was enabled) like this one - https://ci.appveyor.com/project/okonet/lint-staged/build/1.0.79/job/c1u2k57uwn2dn91o. I had also got `EPERM` issues on my local Windows system quite a few times. Which was why I had to switch to yarn. I'll concede that it has been some time since I last used npm so the issues might have been solved. However, considering yarn works quite well, I haven't had the motivation to try npm again(there are other issues when trying to use it with [`nvm-windows`](https://github.com/coreybutler/nvm-windows)).

> I'm not aware of outstanding reasons not to commit the lockfile

Let's say you declared the dependency `pkg-a: ^1.1.0` and this was locked to `1.1.3`. If a newly released version, say `1.2.0`, breaks any of the tests, only the end-user will be able to see it because when we run any test in CI, yarn will install `1.1.3` and not `1.2.0`. But this is an edge case and I don't think this is a real risk for `cosmiconfig`.

> I am unsure about ``npm``, but ``yarn``  according to [yarnpkg.com/blog/2016/11/24/lockfiles-for-all](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/):  _"Lockfiles within your dependencies will be ignored"_.

Here's a relevant article - https://jpospisil.com/2017/06/02/understanding-lock-files-in-npm-5.html. `package-lock.json` present in dependencies are ignored similar to `yarn.lock`.

olsonpm:
> But this is an edge case and I don't think this is a real risk for cosmiconfig
Yeah and without something like greenkeeper too, it'd be lucky if we even caught updated dependency incompatibilities

chrisblossom:
Which version of yarn does this repo use? When running ``yarn`` with the most recent version ``1.5.1`` it updates the ``yarn.lock`` file.

> Yarn is fully deterministic as long as all your teammates are using the same Yarn version.

When using yarn, everyone needs to use the same version according to https://yarnpkg.com/blog/2017/05/31/determinism/

This might be another reason to remove ``yarn.lock``?

davidtheclark:
> This might be another reason to remove yarn.lock?

I'm fine removing Git-ignoring `yarn.lock` if others want to.

sudo-suhas:
I don't have a strong opinion but if I had to vote, I'd keep the lock file.

davidtheclark:
Looks like the `--pure-lockfile` option might be relevant here. If you're not actually updating dependencies, you could use `yarn --pure-lockfile` to use the lockfile as the source of truth, ignoring `package.json`.

PR #120, opened 2018-03-26
[WIP] Expose sync functions instead of a sync option, remove load format

### LINKED ISSUE #112: Provide searchSync and loadSync functions, instead of the sync option

davidtheclark:
Closes #112. **Not quite finished yet, though.**

**The principle motivating these changes is that every function (public and internal) should be *either* sync *or* async.** So the bulk of the refactoring is to remove `sync` options that make a single function alternately sync *and* async. 

The goal is that each function will be more simple and type-checkable. (I think contrasting the new loader functions with the old ones really illustrates the benefit.)

**Public API changes:**

- No more `sync` option.
- Instead of the `sync` option, the explorer exposes `searchSync` and `loadSync`; and the non-`*Sync` alternatives are always async.
- No more `format` option (as discussed in https://github.com/davidtheclark/cosmiconfig/pull/107#issuecomment-372341095). (This is unrelated to sync/async, except that it made sense to remove it now instead of doing work to incorporate it into the refactor.)

**Internal changes:**

- `getDirectory` is async, `getDirectory.sync` is sync.
- `loadDefinedFile` is async, `loadDefinedFile.sync` is sync.
- Loader functions are consolidated into `loader.js`. Each file type has separate sync and async loaders.
- `loaderSeries` and `loaderSeries.sync` essentially replace `funcRunner`.
- Parser functions are consolidated into `parser.js`. This is kind of unrelated to the sync/async crusade, since all the parsers are sync; but I think it's the right thing to do (we were repeating that logic) and it made the sync/async refactor better.
- All these changes allow Flow to do more and better work, so the branch includes some Flow type refinements.
- The only test changes are to use the new `searchSync` and `loadSync` functions, instead of the `sync` option. 

**What I'm still dissatisfied with 🤔:**

- The `search`/`searchSync` and `load`/`loadSync` function pairs each repeat ~20 lines of important logic with very slight variations. Does anybody have any ideas about how to reduce that repetition *while still maintaining the principle state above* (each function is either sync or async, not both) ❓❓❓

**Still to do:**

- Documentation changes.
- Changelog entry.
- Increase code coverage back to 100% (all existing tests still pass, but there seem to be some uncovered lines in the refactored code).


davidtheclark:
To be clear: I'm not 100% convinced about this refactor yet. I'd be very curious to hear what some of the contributors newer to the codebase think (@olsonpm @chrisblossom). Of course also @sudo-suhas 😄 I don't know if the principle of separating sync & async functions is a firmly established one that other engineers value, or if I'm overemphasizing its benefits.

If people disagree with the refactor and the breaking change, I could take some of the less disruptive refactoring that I like from this branch and try again.

davidtheclark:
Some precedent for favoring separate sync and async functions: http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony

sudo-suhas:
I was in favor of splitting the function because the API for the end user would be explicit and easier to understand/reason about. However, I too am somewhat dissatisfied with the code repetition and some of the other changes. For one thing, we now have a lot of functions(15 in `loader.js`). Furthermore, I am not fond of `loaderSeries` and it's sync counterpart. I was not expecting it to have the logic for calling the next loader and `ignoreEmpty`. Here's how I would prefer writing it:

```js
async function funcRunner(
  funcs: Array<(CosmiconfigResult) => null | Promise<CosmiconfigResult>>,
  returnEarly: (CosmiconfigResult) => boolean
): Promise<CosmiconfigResult> {
  let result = null;
  for (const fn of funcs) {
    result = await fn(result);

    if (returnEarly(result)) return result;
  }

  return result;
}

funcRunner.sync = function funcRunnerSync(
  funcs: Array<(CosmiconfigResult) => CosmiconfigResult>,
  returnEarly: (CosmiconfigResult) => boolean
): CosmiconfigResult {
  let result = null;
  for (const fn of funcs) {
    result = fn(result);

    if (returnEarly(result)) return result;
  }

  return result;
};

```

This uses unsupported feature async-await but we could use [`co`](https://github.com/tj/co) to achieve nearly the same thing. There's also a flow error which I don't know how to fix - `[flow] function type (Callable signature not found in object literal)`.

For the functions `searchDirectory` and `searchDirectorySync`, I want to say let's use a `createLoader` which comes with async/sync versions as appropriate **but** that would be going back to what we had earlier.

davidtheclark:
> For one thing, we now have a lot of functions

I don't understand this concern.

> I am not fond of loaderSeries and it's sync counterpart. I was not expecting it to have the logic for calling the next loader and ignoreEmpty.

I don't understand this one, either. We had a problem: too much code in scattered places needed to independently address `ignoreEmpty`. The solution here is to consolidate that into one place. Since `ignoreEmpty` pertains to how we deal with a series of loaders — whether we move from one to the next or not — it seemed appropriate to address it in `loaderSeries`.

I'm fine with using a for-loop instead of `reduce`! I was just carrying over some of `funcRunner`, but you're right a loop would be a little more efficient.

> This uses unsupported feature async-await but we could use co to achieve nearly the same thing. 

We could just use Promises.

davidtheclark:
Added a new commit that reduces some repetition in `createExplorer` by create a `cacheWrapper` function to handle repeated cache getting and setting logic.

Also used a for-loop for `loaderSeries.sync`. (Can't do that in the async function, because of Promises — now I understand why async-await seemed important. If we can figure out a simple way to return early from `loaderSeries` (async), great; but I don't think it's worth adding complexity, especially not adding a dependency or build step).

olsonpm:
I'm taking a look at it now - mind explaining why you separated the a/sync caches?  previously they were shared.

davidtheclark:
> mind explaining why you separated the a/sync caches? previously they were shared.

@olsonpm Same base reason for separating sync/async functions: each now has one responsibility (it stores *either* async *or* sync results, not both), instead of instead of one cache having both sync and async responsibilities.

olsonpm:
Sounds good - just seems to me that the purpose of the cache is to prevent unnecessary fs calls - which the sync and async versions will be sharing the same of.  Personally I felt even keeping the load and search caches separate seemed odd given that premise - though i'm sure there are tons of use-cases I'm naively disregarding.

davidtheclark:
> just seems to me that the purpose of the cache is to prevent unnecessary fs calls - which the sync and async versions will be sharing the same of

The same fs-call prevention is happening as before (this is tested). It was never shared between sync and async calls, because the the async cache stores Promises resolving with results while the sync cache stored the actual result. I don't think we need to build caches to accommodate a user switching between sync and async, either.

chrisblossom:
The internal API doesn't have to match the external API. Also, the internal API doesn't have to be all or nothing in regards to separating ``async`` and ``sync`` functions.

From this discussion, I think first you should decide if you want to have the proposed ``*Sync`` or current ``sync: true`` option public API.

It seems like the real goal of this change is to refactor the code base to make it easier to understand / extend. I think this PR goes a long way to do that, particularly with ``loader.js`` and ``parser.js``, but too far in completely removing the ``sync`` option internally.

davidtheclark:
Thanks all for sharing your thoughts! 

Sounds like everyone agrees that the public API change is good. I guess I'd stress that all function APIs are *for* somebody: the APIs of internal functions are for the library devs. So the principle should apply in pretty much the same way to public and non-public functions. But I do agree that it's not a good idea to repeat some of the logic that is really key to the program, and if there's no better way to avoid such repetition in `createExplorer`, maybe writing a couple of internal functions that handle both sync and async is the better choice there.  I will try another commit that does this.

I don't think I'm convinced by concerns about the *quantity* of functions or *verbosity* the code, though. I've repeatedly found that having *more, simpler* functions has paid off better than *fewer but more complex* functions.

olsonpm:
> I don't think I'm convinced by concerns about the quantity of functions or verbosity the code

This, like any code-style opinions, is a value judgement.  It's fair and probably reflects your personal experience - but I would argue that verbosity tends to get in the way of the logic.  Even just type declarations are noise to me whereas others really prefer them.  In the end the most important part is the external api which majority of your users will interact with.

davidtheclark:
😠😠😠It really seems like there's no good way to have meaningful Flow annotations (with meaningful errors) and also write some functions that have two signatures, one sync and one async. Uuuugh, I'm really frustrated that there don't seem to be any good choices here, only bad ones. I'm entertaining the thought of jettisoning Flow completely, but its complaint does accurately indicate that the code's types are unclear (in a function like this); but the unclarity of the types is a deliberate move in order to avoid repetition of key logic. Puke.

davidtheclark:
Figured it out with the latest commit 😅 

For some reason I'd forgotten that I could selectively use annotations like `any` and `Function` to skip type checking in certain contexts. As far as I know, that's the best we can do in this situation. So I created a couple of functions to encapsulate logic for both sync and async cases; annotated those in a way that skipped over type checking them; and now things work. Those functions are `getSearchDirectoryLoaderSeries`, `getLoadResult`, and `getRcLoaderSeries`.

In the end, I think I'm pretty satisfied with where this ended up. A lot of functions that each encapsulate some repeated functionality but most of which are as simple as can be, with like 0-1 levels of branching.

I'll leave this up for a couple of days to see if there are any more ideas or suggestions. Then I'll plan to update docs for the public-facing API change and merge.

olsonpm:
Awesome - Once you merge I'll be able to update those hanging pr's in stylelint and prettier.  I didn't foresee the further refactoring work you ended up doing ha.

sudo-suhas:
> > For one thing, we now have a lot of functions

> I don't understand this concern.

Yea, this one is a little hard to justify for me too. But it really boils down to this. Why do we have so many functions?

What we used to have:
image

We have a total of 53 functions now:
image

While this is a crude way to look a things, I'd still like to pause and ask if there is something we could do a bit differently while achieving our target of separate sync and async functions.

> I don't understand this one, either. We had a problem: too much code in scattered places needed to independently address `ignoreEmpty`. The solution here is to isolate that logic in one place. Since `ignoreEmpty` pertains to how we deal with a series of loaders — whether we move from one to the next or not — it seemed appropriate to address it in `loaderSeries`.

> I'm fine with using a for-loop instead of `reduce`! I was just carrying over some of `funcRunner`, but you're right a loop would be a little more efficient.

My apologies again for the lack of clarity. My main intention behind this suggestion was not to use `for` instead of `reduce`. What I was trying to get at was passing `returnEarly` as a parameter so that that logic would be at the call site. As to why, I see the current implementation as separation of important logic into interweaving functions. The runner has to be dumb IMHO. I agree with the 'isolate `ignoreEmpty` logic part' but just not in the way it is implemented currently.

> I don't think it's worth adding complexity, especially not adding a dependency or build step

While this is not very important, I'd still like to discuss it a little if you don't mind. Using `co`, to me at least, would actually be simplifying the code we have to write because it can be written in a seemingly synchronous style. For adding the dependency, why not? `co` itself has no dependencies and the lib is ~200 LOC including comments, newlines etc. As for the build step, we already have one 😉 but we wouldn't need it even if we wanted to use `co`.

sudo-suhas:
@davidtheclark What do you think would be the pros and cons of using the previous version of `funcRunner`? Because as far as I can see, it could cut down the number of functions by nearly half. Is type information of returning specifically `Promise<CosmiconfigResult>` or `CosmiconfigResult` the main motivation behind the similar functions for sync and async?

sudo-suhas:
I like the `cacheWrapper` implementation. very nice.

davidtheclark:
Thanks @sudosuhas! 

> Why do we have so many functions?

The sheer quantity of functions doesn't bother me at all, and I really don't think it's a point of discussion. We'll have to be more specific. 

For example, I think a flaw in the previous code was that it imported parsing libraries like `require-from-string` and `js-yaml` in *several* places, instead of abstracting out the functionality of "parsing JS" and "parsing YAML" to ensure we do it the same way every time. By creating `parseJs` and `parseYaml` functions, we're relieving a few other functions of the responsibility of knowing *how* to parse JS and YAML. If we later change our parsing method, we do it in one place; if a dependency update means we need to adjust our invocation, we do it in one place; if we want to mock that method for testing, it's easy; etc.

I doubt anyone disagrees with *that* example (?). The move I think you have a problem with is splitting out sync and async functions, which turns one function into two. The intention is that each of those two functions will have a cleaner function signature — which I believe is the same code design principle that gives us `fs.readFile(..)` and `fs.readFileSync(..)` instead of `fs.readFile(sync: boolean, ..)`, or even `fs.doThing(action: 'readFile' | 'writeFile' | .., sync: boolean, ..)`.

Do you disagree with that intention? If so, I'm curious why. I've tried reading things and asking people about this, and the opinions I've gathered seem to be unanimously in favor of cleaner function signatures and the separation of sync and async code. (This also addresses your question about the "main motivation behind the similar functions for sync and async".)

> the previous version of funcRunner ... could cut down the number of functions by nearly half

Again, I don't think the sheer quantity of functions really matters?

> What I was trying to get at was passing `returnEarly` as a parameter so that that logic would be at the call site.

The intention was precisely to pull logic out of the call site, so it wouldn't have to be repeated in multiple call sites. I don't know why the "runner has to be dumb" if it's not a *generic* runner, but is instead specifically targeted at running *a series of loaders*. We don't have any other different purposes for a runner, so I don't think we need to make it more generic.

Re `co`: I guess I don't understand how `co` would help. Is your idea that generators might help clean up the functions that need to share sync and async responsibilities? Maybe I need an example? (If the goal is just to use generator syntax with `yield` instead of regular Promises, I don't think I understand the benefit?) I'm totally open to suggestions about those few dual-signature sync-async functions! (I don't like them at all, really, which is the reason I tried to make the *rest* of the codebase *less* like them, with cleaner function signatures. The few that are left seemed like a necessary evil.)

sudo-suhas:
@davidtheclark Please give me time till tomorrow to reply to this.

sudo-suhas:
> For example, I think a flaw in the previous code was that it imported parsing libraries like `require-from-string` and `js-yaml` 

I am not arguing against the `parseX` functions. Apologies again for the vague feedback on that.

> The move I think you have a problem with is splitting out sync and async functions, which turns one function into two. 

Actually this is not entirely true. The points of consideration are
- Is important logic repeated? You already addressed this in your last commit.
- Strongly typed, slightly duplicate functions are present but the type information is lost/not being used? I feel this is the case with the functions in `loader.js` due to the way it is being used in `createExplorer.getSearchDirectoryLoaderSeries`, `loader.getRcLoaderSeries`. For this, I have created a gist with a refactor - https://gist.github.com/sudo-suhas/fd4011b8c9f4ea7f333ce21e5a1e99eb. There are some flow typing errors though.

> The intention was precisely to pull logic out of the call site, so it wouldn't have to be repeated in multiple call sites..

It is possible to create a shared function which could be passed so it need not be repeated. But I concede. What you said makes sense and the feedback does not merit further discussion.

> Re `co`: I guess I don't understand how `co` would help. 

It was a theoretical question along the lines of "Why would not want to use `co` in a hypothetical project". Also, adding `co` does not make anything drastically better for us. The only place I would have liked to use it is in `loaderSeries`. 

> If the goal is just to use generator syntax with yield instead of regular Promises, I don't think I understand the benefit?

Using `co`, `yield` and generator functions allows to write something similar to `async-await`. Eventually, it would be possible to switch to `async-await` itself.

davidtheclark:
I want to go with what's here for now. Thanks for the loader suggestion, @sudo-suhas, but it ends up creating a lot more functions that are both sync and async, which is what I'm trying to avoid. Future refactoring is always a possibility, and we can do that in follow-up PRs.

PR #121, opened 2018-03-28
create a flow definition file in build script

saadq:
This uses the `flow gen-flow-files` command to create an `index.js.flow` file and include it in `dist`. With this file, flowtype users of the `cosmiconfig` package will get the type information you included in your project.

screen shot 2018-03-28 at 8 49 52 am

A definition file (`index.js.flow`) like this will get created:

```ts
// @flow


declare module.exports: (moduleName: string, options: {cache?: boolean, configPath?: string, format?: "json" | "yaml" | "js", js?: string | false, packageProp?: string | false, rc?: string | false, rcExtensions?: boolean, rcStrictJson?: boolean, stopDir?: string, sync?: boolean, transform?: (_: ?Object) => ?Object}) => {clearCaches: () => void, clearDirectoryCache: () => void, clearFileCache: () => void, load: (searchPath: string, configPath?: string) => Promise<?{config: any, filepath: string}> | ?{config: any, filepath: string}};
```

Here is a prettified version if you'd like to verify:

```ts
// @flow

declare module.exports: (
  moduleName: string,
  options: {
    cache?: boolean,
    configPath?: string,
    format?: 'json' | 'yaml' | 'js',
    js?: string | false,
    packageProp?: string | false,
    rc?: string | false,
    rcExtensions?: boolean,
    rcStrictJson?: boolean,
    stopDir?: string,
    sync?: boolean,
    transform?: (_: ?Object) => ?Object,
  }
) => {
  clearCaches: () => void,
  clearDirectoryCache: () => void,
  clearFileCache: () => void,
  load: (
    searchPath: string,
    configPath?: string
  ) =>
    | Promise<?{ config: any, filepath: string }>
    | ?{ config: any, filepath: string },
};
```

If you'd rather not rely on the script to generate the type definition, we could create an `index.js.flow` file in `src` and just copy that over in the `build` script. Another advantage to that would be that it would then be possible to write a test for that libdef.

Let me know if you have any comments on this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "test": "jest --coverage",
     "test:watch": "jest --watch",
     "coverage": "jest --coverage --coverageReporters=html --coverageReporters=text",
-    "build": "flow-remove-types src --out-dir dist --quiet",
+    "build": "flow-remove-types src --out-dir dist --quiet && flow gen-flow-files src/index.js --out-dir dist",
     "prepublishOnly": "npm run build"
   },
   "lint-staged": {


saadq:
On second thought, I think it actually would be better to do this as a separate `index.js.flow` file... there are some type refinements that can be manually be done that would really help out. Currently, Flow will complain if you do

```js
var cosmiconfig = require('cosmiconfig');

var explorer = cosmiconfig('example', {});
var result = explorer.load()

result.then((result) => {
    
})
```

Because the type of `explorer.load()` is not currently guaranteed to be a Promise. So you currently would have to do:

```js
var cosmiconfig = require('cosmiconfig');

var explorer = cosmiconfig('example', {});
var result = explorer.load()

if (result instanceof Promise) {
  result.then((result) => {
    
  })
}
```

to get it to typecheck.

If the `index.js.flow` file was written manually, we could include multiple declarations for `load()` so that Flow would know that it returns a `Promise` if `sync: true` isn't passed.

I'll wait for any comments by you guys before I do this though.

sudo-suhas:
@saadq Do take a look at some of the changes to typings in the `sync_functions`(#120) branch.

sudo-suhas:
@saadq Is this PR still required and is it okay to merge?

@davidtheclark Could you please take a look?

saadq:
I actually haven't been keeping up with any changes there may have been so I'm not sure if it's currently okay to merge – will have to check.

But I would like to first hear if you guys would prefer a build script or a manual `index.js.flow` file with a test for it.

sudo-suhas:
Please do check. I'd also like to hear your opinion on whether `index.js.flow` is still necessary after the recent changes.

chrisblossom:
I don't think using ``flow gen-flow-files`` is a good idea. Not only is it experimental and lots of known issues, there are talks about [deprecating the feature entirely](https://github.com/facebook/flow/issues/5871) (it hasn't been worked on/updated since release). I think using [``flow-copy-source``](https://github.com/AgentME/flow-copy-source) is a more reliable / better supported / future proof solution.

saadq:
Yep, I totally agree that `flow-copy-source` or other solutions would be a lot more preferable. Sorry about my inactivity, but I'm going to close this for now to lessen GitHub noise for the project and maybe come back to it at some point when I can (if someone else doesn't get to it!).

Issue #119, opened 2018-03-23
Feat: use JSON5 instead of JSON.parse for friendly JSON syntax


imcuttle:
[json5](https://github.com/json5/json5)

```js
{
  // comments
  unquoted: 'and you can quote me on that',
  singleQuotes: 'I can use "double quotes" here',
  lineBreaks: "Look, Mom! \
No \\n's!",
  hexadecimal: 0xdecaf,
  leadingDecimalPoint: .8675309, andTrailing: 8675309.,
  positiveSign: +1,
  trailingComma: 'in objects', andIn: ['arrays',],
  "backwardsCompatible": "with JSON",
}
```

olsonpm:
i think there's work toward using `require` instead of parsing the file as a string.  We'll then allow consumers to pass their own require function.  This is mainly to support es modules but your use-case will be covered here as well.

Anyway we can leave this issue open till that's finished

sudo-suhas:
@olsonpm That would apply for JS file only right? Or are we allowing override for other file types as well?

olsonpm:
why we wouldn't require json files?

sudo-suhas:
Because the errors in the default case would not be very helpful. Which is why we use [`parse-json`](https://github.com/sindresorhus/parse-json) right now.

olsonpm:
but requiring would be a lot more friendly toward parsing other flavors of json while keeping the options fewer and thus api cleaner.  How would you choose to handle other flavors of json?

sudo-suhas:
That hasn't really come up until now. Moreover, YAML allows greater flexibility for defining a config file including comments, skipping double-quotes etc. And as I mentioned, I think the default case is more important.

olsonpm:
Yeah I guess I just figured we would require given davidtheclark introduced it as the mechanism for parsing js files - it seems natural to me to also require json files e.g. `const { version } = require('./package.json')`.  Personally I would find json5 more palatable than yaml given it's more familiar.  I'm not a fan of whitespace sensitive syntaxes, and although I understand why they didn't support comments in vanilla json, comments are extremely useful in a lot of circumstances.

@davidtheclark - any thoughts on whether to support json5? and what implementation you'd prefer.

davidtheclark:
I was not planning on using `require` for JSON files. I don't think there's a need.

[...]

PR #124, opened 2018-04-13
remove javascript from extensionless parsing

chrisblossom:
I put this as a PR to show that as of right now the [``tryAllParsing`` "feature"](https://github.com/davidtheclark/cosmiconfig/blob/143ff724d2a791316f68c6298d0d43eda67ee823/src/loadDefinedFile.js#L28-L30) of an unknown extension is broken. There weren't any successful tests, so here are some that __should__ work (but fail).

https://github.com/davidtheclark/cosmiconfig/blob/143ff724d2a791316f68c6298d0d43eda67ee823/src/loadDefinedFile.js#L57-L63
Initially I thought just swapping these functions would solve the issue, but instead ``require`` and ``require-from-string`` will return ``{}`` when presented valid yaml, such as ``yaml: true``.

On the flip side, the yaml parser will return ``{ 'module.exports = { js': 'true }' }`` when parsing a js file.

Additionally I do not think the try all parsing behavior is a good thing to do. It is highly error-prone, and more than likely there was a mistake made and an error should be thrown.

I propose to remove this catchall of unknown file types. Thoughts?

FYI, I think this is the last thing I'm tackling for https://github.com/davidtheclark/cosmiconfig/issues/109. I would have missed this if it weren't for flow because there aren't any successful tests.


davidtheclark:
> More than likely if the user is getting here, there was a mistake made and there should be an error thrown.

@chrisblossom: I believe the typical use case is to have a config file without an extension that can be parsed as either YAML or JSON. I think that feature is worth keeping, right? If it's too hard to also fit in JS parsing, I think we could definitely drop that attempt — say that extensionless files can only be loaded if they are in JSON or YAML format. This is already how `search` behaves, I believe.

chrisblossom:
I am not sure about YAML (no real experience, all of the YAML configs I've seen personally have an extension), but I can see your point with JSON (examples being ``.babelrc`` and ``.eslintrc``). Although I do not think the YAML parser should be responsible for parsing extensionless JSON, which it currently is.

chrisblossom:
I force updated this PR to make the requested changes. Should be ready to be merged unless more changes requested / not wanted. Please review.

For reference, original PR: https://github.com/chrisblossom/cosmiconfig/commit/3e0047f9e3d8be8523abfc6399c584051bb73591.

davidtheclark:
> Although I do not think the YAML parser should be responsible for parsing extensionless JSON

Thanks — this forced me to remind myself of the details here ...

YAML parsers understand JSON because valid JSON is valid YAML — [the spec says](http://yaml.org/spec/1.2/spec.html#id2759572) "every JSON file is also a valid YAML file". So I believe it's fine to use the YAML parser to parse extensionless rc files that happen to contain JSON.

Your understandable concern, though, is the reason we have the `rcStrictJson` option: for those who want to ensure that extensionless rc files for their program are *only* valid JSON, not JSON-plus-some-YAML (which is actually just YAML, I guess).

So if the user has set `rcStrictJson: true`, we should *only* use the JSON parser, not the YAML parser at all. While if they have *not* used that setting, we can just use the YAML parser.

That's how we're doing it in `search` when we `loadRcFileWithoutExtensions`: https://github.com/davidtheclark/cosmiconfig/blob/143ff724d2a791316f68c6298d0d43eda67ee823/src/loader.js#L83-L90. I guess we neglected `load`.

So I think that [the change here to try JSON first before YAML](https://github.com/davidtheclark/cosmiconfig/pull/124/files#diff-548f96d3487b36e2d92e138c870c529bR58) isn't necessary, because it doesn't actually get us anything (e.g. more helpful error messages). If your JSON is not quite strict enough and the JSON parser errors, we're still going to try the YAML parser anyway, right? 

*Instead*, what we should do is make `load` pay attention to the `rcStrictJson` option, as `search` does, and use `parseJson` if it's on. (We could include that here or take care of it in another commit.)

That's my understanding, at least. Make sense?

chrisblossom:
Definitely makes sense.

> So I think that the change here to try JSON first before YAML isn't necessary, because it doesn't actually get us anything (e.g. more helpful error messages). If your JSON is not quite strict enough and the JSON parser errors, we're still going to try the YAML parser anyway, right?

https://github.com/davidtheclark/cosmiconfig/blob/143ff724d2a791316f68c6298d0d43eda67ee823/src/loadDefinedFile.js#L19-L20

I think parsing should be explicit and consistent. Currently rightfully so, ``JSON`` has its own parser. I would expect that the ``JSON`` parser be used for all ``JSON`` parsing, not sometimes the ``YAML`` parser depending on how it is loaded. In other words, ``.babelrc`` written in ``JSON`` and ``.babelrc.json`` should be parsed the exact same way.

Another reason why it should be consistent is for things like https://github.com/davidtheclark/cosmiconfig/issues/119 requesting [JSON5](https://github.com/json5/json5) support. I do think this is a valid request, even [babel](https://github.com/babel/babel/blob/a6df92f2459d410dec878b682ef5e2eaa6789cbf/packages/babel-core/package.json#L40) uses it to parse their ``JSON`` configs. Maybe #109 should allow custom ``loadJson`` and ``loadYaml`` loaders too?

Thoughts?

> So if the user has set rcStrictJson: true, we should only use the JSON parser, not the YAML parser at all. While if they have not used that setting, we can just use the YAML parser.

I can add this to the PR if wanted.

davidtheclark:
Thanks again, @chrisblossom. I'll merge this and we'll move on from here.

> In other words, .babelrc written in JSON and .babelrc.json should be parsed the exact same way.

I don't know of a way to do this, unless you the user use the `rcStrictJson` option to disallow extensionless config files from being interpreted as YAML. Besides, the YAML parser's treatment of JSON has been perfectly consistent and reliable.

> Maybe #109 should allow custom loadJson and loadYaml loaders too?

Maybe we should add an ability to add your own arbitrary files to the list that gets attempted in each directly — that way when someone in the future requests XML or INI or CSV or whatever support, it will already be available to them.

Issue #126, opened 2018-04-15
Proposal for a new v5 options scheme to allow for all kinds of flexibility


davidtheclark:
We've had quite a few requests, some recent and some older, asking for more flexible ways to find, name, and parse files. Here's what I dug up:

- [ES2015 modules](https://github.com/davidtheclark/cosmiconfig/issues/109)
- [Whatever ES2015+ syntax Babel can handle](https://github.com/davidtheclark/cosmiconfig/issues/109)
- [Alternate JS extensions](https://github.com/davidtheclark/cosmiconfig/issues/122)
- [TypeScript](https://github.com/davidtheclark/cosmiconfig/issues/122)
- [JSON5](https://github.com/davidtheclark/cosmiconfig/issues/119)
- [`.config.js` files starting a dot](https://github.com/davidtheclark/cosmiconfig/issues/57), like `.foo.config.js`
- [A subproperty in `package.json` nested under a `config` property](https://github.com/davidtheclark/cosmiconfig/issues/56), like `{"config":{"foo":{..}}}`
- Searching [within a `.config/` subdirectory of each searched directory](https://github.com/davidtheclark/cosmiconfig/issues/50)

We've also had [a discussion](https://github.com/davidtheclark/cosmiconfig/pull/124) about the way extensionless rc files are parsed. We've talked about adding a `parseJs` option (to address the request above for alternate flavors of JS); and, as was inevitable, the suggestion came up for `parseJson` and `parseYaml` options to allow for alternate flavors of *those* languages — which got me anticipating eventual requests to parse XML or INI or yet-unheard-of languages. It also got me looking at the current `rc`, `rcStrictJson`, and `rcExtensions` options, and realizing how conceptually tangled they are.

So, **I'm wondering if there's a way we could set up v5 to allow for *any filenames*, searched for in *any order*, and parsed with *any parsers*.**

Here's what I've come up with:

What if we do away with most of the existing options (`packageProp`, `rc`, `js`, `rcStrictJson`, and `rcExtensions`), and instead have a flexible `searchScheme` option. 

- `searchScheme` is an array representing an ordered list of things to search for in each directory.
- Each item in the array is either an object or a string.
- Object items can have the following properties:
  - `filename`: The filename of a file to search for, relative to the directory being searched; e.g. `.stylelintrc`, `.stylelintrc.json`, `stylelint.config.js`, `package.json`. Could even include a directory if you wanted to search in a `.config/` subdirectory of every searched directory.
[...]

PR #127, opened 2018-04-15
Use require for js and replace mocks with real file tests

chrisblossom:
This PR replaces ``require-from-string`` with ``require`` and replaces filesystem mocks with real file tests. @davidtheclark if you aren't happy with this implementation, feel free to close.

Working from this commit I have the majority of the ``loadJs`` feature added, but it looks like that will be superseded with #126. I'll hold off with that feature until #126 is decided on, unless you think otherwise?


src/loader.js
```diff
@@ -12,7 +12,20 @@ function fileContentToResult(
...
+   * https://stackoverflow.com/a/10805292
+   */
+  const contentWithoutNewLines = content.replace(/\r?\n|\r/, '');
+  if (contentWithoutNewLines === '') {
+    return {
+      filepath,
+      config: undefined,
+      isEmpty: true,
+    };
+  }
```
chrisblossom: When switching to real file tests I found that the content wouldn't return an empty string because of new lines. Pretty sure this was a bug.

test/util.js
```diff
@@ -1,55 +1,86 @@
...
+class TempDir {
+  dir: string;
 
-exports.mockStatIsDirectory = function mockStatIsDirectory(result) {
-  const stats = {
-    isDirectory: () => result,
-  };
+  constructor() {
+    const tempDir = fs.realpathSync(os.tmpdir());
+    const parent = parentModule();
```
sudo-suhas: Hey @chrisblossom, could you please explain the significance of these 2 lines?

test/util.js
```diff
@@ -1,55 +1,86 @@
...
-  } else {
-    return jest
-      .spyOn(fsMock, 'readFile')
-      .mockImplementation(makeReadFileMockImpl(readFile));
+  createDir(dir: string) {
+    const dirname = this.absolutePath(dir);
+    makeDir.sync(dirname);
+  }
+
+  createFile(file: string, contents: string | Object) {
```
sudo-suhas: Contents is always a string right? `${contents}` would not work well either if it was an object.

sudo-suhas:
> This looks great! Thanks @chrisblossom.

Agreed 👍!

test/util.js
chrisblossom: Yes, good catch! `Object` was added from a previous implementation that I forgot to remove.

test/caches.test.js
```diff
@@ -1,456 +1,439 @@
...
+  temp.deleteTempDir();
 });
 
 describe('cache is not used initially', () => {
-  const searchPath = absolutePath('a/b/c/d/e');
-  const checkResult = (readFileMock, result) => {
-    util.assertSearchSequence(readFileMock, [
+  const searchPath = temp.absolutePath('a/b/c/d/e');
+  const checkResult = (readFileSpy, result) => {
+    const searchPath = temp.getSpyPathCalls(readFileSpy);
```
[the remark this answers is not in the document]
chrisblossom: Do you mean only in ``caches.test.js`` or for all variables named ``searchPath`` in all tests files?

src/loader.js
chrisblossom: ``trim`` worked perfectly. I [added a test](https://github.com/davidtheclark/cosmiconfig/pull/127/files#diff-7a54a2eed8c64ebde960aa063ef05ba3R476) to confirm.

test/caches.test.js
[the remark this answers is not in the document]
davidtheclark: I meant in all the places this pattern (`getSpyPathCalls`) is used.

test/util.js
chrisblossom: I've [added some comments](https://github.com/davidtheclark/cosmiconfig/pull/127/files#diff-d8642304b1d7b6801be66d60ee3cbc5bR13). Please let me know if that clears things up.

test/caches.test.js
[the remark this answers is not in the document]
chrisblossom: Switched them all to ``filesChecked``.

chrisblossom:
I've updated the PR to reflect the requested changes and fixed the tests. I reverted the updated Jest change so it will work on node 4 and [updated `getSpyPathCalls` to normalize windows directory separators](https://github.com/davidtheclark/cosmiconfig/pull/127/files#diff-d8642304b1d7b6801be66d60ee3cbc5bR63).

Also, `v5` might be a good time to drop `node 4` support. [The end of life is scheduled for the end of this month](https://github.com/nodejs/Release#release-schedule). People wanting to support `node 4` could continue to use `cosmiconfig 4` as they are currently. Maybe should be bought up as a separate issue?

Regarding this PR, what do you think about replacing all files in ``fixtures`` to use the same ``TempDir`` test pattern?

davidtheclark:
> Also, cosmiconfig v5 might be a good time to drop node 4 support. The end of life is scheduled for the end of this month. People wanting to support node 4 could continue to use cosmiconfig v4 as they are currently. Maybe this should be bought up as a separate issue?

Yeah, a separate, focused issue makes sense. Because cosmiconfig is low-level, I'd rather that cosmiconfig not force higher-level tools, its users, to stop supporting Node 4 (if those higher-level users have not already stopped). I know that Prettier still supports Node 4, for example. So unless absolutely necessary (and I don't think it should be), I'd prefer to keep supporting Node 4 until we're pretty confident that the bulk of cosmiconfig users have already dropped support.

> Regarding this PR, what do you think about replacing all files in fixtures to use the same TempDir test pattern?

Yeah, that's a good idea 👍

chrisblossom:
`fixtures` removed and the rest of the tests migrated to `TempDir`. [Found another place](https://github.com/davidtheclark/cosmiconfig/pull/127/files#diff-548f96d3487b36e2d92e138c870c529bR12) where content needed to be checked for whitespace too.

Anything else that I should do for this PR?

test/util.js
```diff
@@ -1,55 +1,101 @@
...
+
+    const result = calls.map(call => {
+      const filePath = call[0];
+      const relativePath = path.relative(this.dir, filePath);
+
+      /**
+       * Replace Windows backslash directory separators with forward slashes
+       * so expected paths will be consistent cross platform
+       */
+      const normalizeCrossPlatform = relativePath.replace(/\\/g, '/');
```
sudo-suhas: Can we use [`slash`](https://github.com/sindresorhus/slash) instead?

test/util.js
```diff
@@ -1,55 +1,101 @@
...
+    // create directory
+    makeDir.sync(this.dir);
 
-exports.assertSearchSequence = function assertSearchSequence(
-  readFileMock,
-  searchPaths,
-  startCount
-) {
-  startCount = startCount || 0;
+    (this: any).absolutePath = this.absolutePath.bind(this);
```
sudo-suhas: Would it be better to use and return a closure instead of the class? We could then avoid having to bind the methods.

test/util.js
chrisblossom: The idea is to have a consistent directory name for each test for things like ``jest --watch`` and other third party tools to play nicely.

Also, because ``tmp`` doesn't use ``realpathSync``, it will break any tests [using ``process.chdir``](https://github.com/davidtheclark/cosmiconfig/pull/127/files#diff-4b7ed0b193d73a1b1fa66c79c737b673R463).

test/util.js
chrisblossom: I use ``slash`` in my own projects and considered adding it here, but I don't think it is worth (having to track updates etc) adding a new package for one line of code. ``slash`` has [guards against unicode and extended paths](https://github.com/sindresorhus/slash/blob/6fd50ff0bcb68812df18d526c0c3d5740fd13382/index.js#L3-L8), which isn't an issue with these tests. Do you disagree?

Also the latest version of ``slash`` doesn't support node v4 so if somehow there was an update (not sure what they could possibly change), we wouldn't benefit from it.

davidtheclark: Yeah, I don't think we need it unless it solves a problem we have.

test/util.js
chrisblossom: Doesn't really matter to me. I think if a binding is required in the class, it is also required in a closure. That said, I can also just remove the bindings because they aren't required with the current implementations. I've just always considered it best practice to keep them in there to guard against anything unexpected happening.

I guess the "trade-off" with a closure would be another level of indentation or having to return the functions.

Just let me know if I should convert it to a closure, remove the bindings, or leave as is.

davidtheclark: I'd also be ok with merging this and creating follow-up issues if we want to refactor. I'm hoping that next week I'll have time to get to work on #126, and I'd love to have these test changes merged in before I launch into it.

test/util.js
sudo-suhas: I see, thanks for explaining.

test/util.js
sudo-suhas: Okay, let's merge this in.

PR #139, opened 2018-05-12
chore: add Node.js 8 and 10

DanielRuf:


sudo-suhas:
Hey @DanielRuf, I noticed that you removed node < 8 from the list of versions to test on. However, we have not dropped support for node 4 yet. Since this is a heavily depended upon package, I think we'd want to allow all packages who depend on cosmiconfig to have some more time before we make the breaking change.

DanielRuf:
Ok I'll add them again even if Node.js 4 reached its EOL.

DanielRuf:
Done.

davidtheclark:
Replaced by #141.

PR #140, opened 2018-05-12
chore: use yarn cache

DanielRuf:


olsonpm:
is there a reason all these chores are separate PRs?  seems they are all meant to be build improvements and could be lumped together.

sudo-suhas:
I am in agreement with @olsonpm. @DanielRuf Would you mind creating a single PR on the lines of `Tweak CI config` for the 3 PRs? You could keep the commits separate.

sudo-suhas:
Also, the caching for yarn can be added on Appveyor as well. Here's an example PR which does so - okonet/lint-staged#378.

DanielRuf:
Will do so later today. Thanks for the feedback.

DanielRuf:
> Also, the caching for yarn can be added on Appveyor as well.

It is there already, see https://github.com/davidtheclark/cosmiconfig/blob/master/appveyor.yml#L17

This is why I did not update the appveyor config in this case.

davidtheclark:
Replaced by #141.

PR #142, opened 2018-05-13
fix code examples in readme

saadq:
Just updated the code examples to use `sync` when appropriate and I removed the `[]` brackets from the `load()` methods because I believe that `loadPath` is required, not optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ If it's a file, the search starts in that file's directory.
 ## explorer.searchSync()
 
 ```js
-const result = explorer.search([searchFrom]);
+const result = explorer.searchSync([searchFrom]);
 ```
 
 Synchronous version of [`search()`].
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -175,7 +175,7 @@ Returns a [result] or `null`.
 ## explorer.load()
 
 ```js
-explorer.load([loadPath]).then(result => {..})
+explorer.load(loadPath).then(result => {..})
```
davidtheclark: If we want to remove the brackets indicating that it's optional (I'm fine with that), let's do it in every place — so also in `search` and `searchSync`.

olsonpm: it is optional in search though.  it's mandatory in load

davidtheclark: Ooooo, right 👍

sudo-suhas:
@davidtheclark Do you think we should do a patch release to update the documentation on npm?

davidtheclark:
Sure, @sudo-suhas 👍

sudo-suhas:
Released `cosmiconfig@5.0.3` :tada:

PR #143, opened 2018-05-16
Allow passing fs instance to createExplorer

### LINKED ISSUE #105: Use a virtual file system
On a [GitHub app](https://developer.github.com/apps) the file system and the cloned repo are not available. Currently `cosmiconfig` browse the filesystem to find the config files, so it cannot be used with a GitHub app.

Would it be possible to pass to the `load` function a `Array` of file `Objects`, which would contains the the file path and the file content as a `String` or a `Buffer`?
So a GitHub app could retrieve the config files as a `String` or `Buffer` and pass them to `cosmiconfig`.

For example:
```js
const explorer = cosmiconfig('mymodule', [
  {path: 'mymodulerc.yml', content: '....'},
  {path: 'dir/mymodulerc.yml', content: '....'},
]);

explorer.load();
// => use config from `mymodulerc.yml`

explorer.load('dir');
// => use config from `dir/mymodulerc.yml`

explorer.load(null, 'dir/mymodulerc.yml');
// => use config from `dir/mymodulerc.yml`
```

sudo-suhas:
- Defaults to `require('fs')`.
- fs instance needs to implement the following methods:
  - `stat`, `statSync`
  - `readFile`, `readFileSync`

Closes #105.

### DIFF
--- a/src/isDirectory.js
+++ b/src/isDirectory.js
@@ -1,0 +1,45 @@
+// @flow
+'use strict';
+
+function isDirectory(fs: FS, filepath: string): Promise<boolean> {
+  if (typeof filepath !== 'string') {
+    return Promise.reject(new TypeError('expected filepath to be a string'));
+  }
+
+  return new Promise((resolve, reject) => {
+    fs.stat(filepath, (err, stat) => {
+      if (err) {
+        // istanbul ignore else
+        if (err.code === 'ENOENT') {
+          return resolve(false);
+        }
+
+        // istanbul ignore next
+        return reject(err);
+      }
+
+      resolve(stat.isDirectory());
+    });
+  });
+}
+
+isDirectory.sync = function isDirectorySync(fs: FS, filepath: string): boolean {
+  if (typeof filepath !== 'string') {
+    throw new Error('expected filepath to be a string');
+  }
+
+  try {
+    const stat = fs.statSync(filepath);
+    return stat.isDirectory();
+  } catch (err) {
+    // istanbul ignore else
+    if (err.code === 'ENOENT') {
+      return false;
+    }
+
+    // istanbul ignore next
+    throw err;
+  }
+};
+
+module.exports = isDirectory;
--- a/src/isDirectory.js
+++ b/src/isDirectory.js
@@ -2,10 +2,6 @@
 'use strict';
 
 function isDirectory(fs: FS, filepath: string): Promise<boolean> {
-  if (typeof filepath !== 'string') {
-    return Promise.reject(new TypeError('expected filepath to be a string'));
-  }
-
   return new Promise((resolve, reject) => {
     fs.stat(filepath, (err, stat) => {
       if (err) {
@@ -24,10 +20,6 @@ function isDirectory(fs: FS, filepath: string): Promise<boolean> {
 }
 
 isDirectory.sync = function isDirectorySync(fs: FS, filepath: string): boolean {
-  if (typeof filepath !== 'string') {
-    throw new Error('expected filepath to be a string');
-  }
-
   try {
     const stat = fs.statSync(filepath);
     return stat.isDirectory();
--- a/README.md
+++ b/README.md
@@ -442,7 +442,7 @@ If you want to throw an error, or if an empty configuration file means something
 
 ### fs
 
-Type:
+Type: Any object that's compatible with node's fs module. Specifically the following must be implemented:
 
 ```js
 interface ErrnoError extends Error {
@@ -464,6 +464,9 @@ interface FS {
   readFileSync(path: string, encoding: string): string;
 };
 ```
+
+Note that only a subset of the [fs API](https://nodejs.org/api/fs.html) is used.
+Also, use of flow is not required for using this config option.
 
 Default: `require('fs')`.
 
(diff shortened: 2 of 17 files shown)


sudo-suhas:
We could add tests which use a virtual file system. Or should we just use a virtual file system for all our tests(instead of creating temp directories)?

olsonpm:
I thought they explicitly wanted real directory tests for `require`, but I'm having trouble finding where that discussion was.

src/isDirectory.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+'use strict';
+
+function isDirectory(fs: FS, filepath: string): Promise<boolean> {
+  if (typeof filepath !== 'string') {
+    return Promise.reject(new TypeError('expected filepath to be a string'));
+  }
```
davidtheclark: Curious: Why do we need this assertion? We aren't making similar assertions elsewhere, right?

sudo-suhas: I referred the `is-directory` implementation and only modified it to return a `Promise`.

davidtheclark: Ah, well I think we should remove this since we have Flow in place and aren't asserting against argument types elsewhere.

sudo-suhas: Alright, done.

davidtheclark:
@sudo-suhas want to add some integration tests against one of the known virtual-file-system libraries?

sudo-suhas:
Okay, will do. Running all the existing tests against both real and virtual fs would be overkill right?

davidtheclark:
> Running all the existing tests against both real and virtual fs would be overkill right?

I think that might be overkill, yes. 😄

sudo-suhas:
Apologies if you got spammed with notifications :sweat_smile:. Was trying to fix Windows errors using appveyor, too lazy to reboot into Windows.

davidtheclark:
Looks great to me 👍 

@sudo-suhas do you feel it's ready to merge?

sudo-suhas:
Yes, I think it is. Would have liked to know what @pvdlg thought of it.

sudo-suhas:
@pvdlg I will wait till tomorrow to hear your opinion. Will merge it in and release as `5.1.0` otherwise.

pvdlg:
Sorry for the late answer. I'm not really knowledgeable about the `cosmiconfig` code so I don't have much of an opinion on the code itself.

Regarding the feature itself that the method that need to be defined are `stat`, `statSync`, `readFile` and `readFileSync`. Is it necessary to define both the sync and async functions? What if use only the async version of `load()`?

In the context of reading the config from the repo within a GitHub app (the example I gave when I opened the issue) I'm not sure how to implement `stat` and `statSync`. Should these function return a result similar to [fs.stat](https://nodejs.org/api/fs.html#fs_fs_stat_path_callback)? How similar (i.e. what fields are necessary)?

Finally it seems surprising for the doc to use Flow. Nowhere else in the doc Flow is used so it's confusing. It might make things unclear for people who don't know Flow and it might give the impression that Flow is required.

olsonpm:
> What if use only the async version of load()?

We've made the decision to split out sync and async functions [back here](https://github.com/davidtheclark/cosmiconfig/pull/120), so that's just part of the api going forward.
<br>

> I'm not sure how to implement stat and statSync

You can think of node's `fs` as the reference implementation to follow.  I'll take a look at how webpack communicated this and see if we can add a note making this clear.
<br>

> Finally it seems surprising for the doc to use Flow

I assume he used flow out of habit, but because javascript doesn't have the concept of interfaces - and what sudo-suhas is exposing is an interface - I can see why using flow to describe the api is fitting.  Again I'll see how webpack communicated this to see if it makes sense to follow suit.

sudo-suhas:
> I don't have much of an opinion on the code itself.

That's alright, I was looking for feedback on the documentation and usability.

> Is it necessary to define both the sync and async functions? What if use only the async version of `load()`?

You could very well not provide methods for the ones you won't be using or throw a not implemented error in them. But in the context of a virtual in-memory fs, it might make more sense to use the sync versions.

> I'm not sure how to implement `stat` and `statSync`.

Wouldn't you use a virtual fs instance created using something like [`memory-fs`](https://github.com/webpack/memory-fs)?

> Should these function return a result similar to `fs.stat`? How similar (i.e. what fields are necessary)?

The [documentation](https://github.com/davidtheclark/cosmiconfig/blob/651f730a3fc766d17c8085d8570fecd3b0f259b6/README.md#fs) answers this. Only the `isDirectory` method needs to be implemented on the `Stats` object which is returned.

> Finally it seems surprising for the doc to use Flow. Nowhere else in the doc Flow is used so it's confusing.

The objective was to express the API. I myself have not used flow much but the type declaration was easy to understand and relatively easy to maintain since it was also present in the [definitions file](https://github.com/davidtheclark/cosmiconfig/blob/651f730a3fc766d17c8085d8570fecd3b0f259b6/flow-typed/defs.js#L46-L63). If you have any suggestions about alternative syntax for expressing the same, I'd be glad to revise it.

> Maybe we could put something like...

I like this suggestion and have updated the documentation for it.

davidtheclark:
This branch is too stale to keep the PR open, and lack of movement suggests lack of interest. Please open a new PR if the change still seems important.

PR #146, opened 2018-05-19
Codecov integration and badge

### LINKED ISSUE #133: Add code coverage badge
This project has 100% code coverage. That's probably a thing worth highlighting in the readme badges — gives potential users some confidence.

davidtheclark:
Closes #133.

I have never hooked up codecov before. I think this should do the trick.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,15 @@
 language: node_js
-node_js:
-  - "4"	
-  - "6"	
-  - "8"
-  - "10"
 git:
   depth: 5
 cache: yarn
 branches:
   only:
     - master
-    - v5
+matrix:
+  include:
+    - node_js: "4"
+    - node_js: "6"
+    - node_js: "8"
+    - node_js: "10"
+    - node_js: "8"
+      script: npm run report-coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,5 @@ matrix:
     - node_js: "4"
     - node_js: "6"
     - node_js: "8"
+      after_success: "npm run report-coverage"
     - node_js: "10"
-    - node_js: "8"
-      script: npm run report-coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,5 @@ matrix:
     - node_js: "4"
     - node_js: "6"
     - node_js: "8"
-      after_success: "npm run report-coverage"
+      after_success: "bash <(curl -s https://codecov.io/bash)"
     - node_js: "10"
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,13 +1,15 @@
...
-  - "8"
-  - "10"
 git:
   depth: 5
 cache: yarn
 branches:
   only:
     - master
-    - v5
+matrix:
```
sudo-suhas: Could you please tell me the significance of using matrix?

package.json
```diff
@@ -17,6 +17,7 @@
     "test": "jest --coverage",
     "test:watch": "jest --watch",
     "coverage": "jest --coverage --coverageReporters=html --coverageReporters=text",
+    "report-coverage": "jest --coverage --coverageReporters=lcov && codecov -t d6ade9c1-c2b4-4ab1-82f4-b2feed929ca4",
```
sudo-suhas: Is that a token for Codecov?

.travis.yml
davidtheclark: Matrix allows us to do different things in different Travis jobs. My understanding is that defining multiple versions of Node is effectively shorthand for defining a Matrix. But in this PR I expanded that into the longhand so I could add another job that performs the same test but sends code coverage results.

package.json
davidtheclark: Yes, that is the project-specific token.

davidtheclark:
> Is there any reason this won't work for us?

That doesn't appear to be the method that is documented and suggested: https://github.com/codecov/example-node

davidtheclark:
I was able to remove the extra job from Travis and use one of the existing jobs to report coverage, because I realized that the existing jobs were already each generating coverage data.

I think the reason to still keep the Matrix, instead of a single `after_success` hook, is because we only need one of the jobs to report coverage, not all of them.

sudo-suhas:
@davidtheclark See https://docs.codecov.io/docs/about-the-codecov-bash-uploader.  Since codecov is a CI specific dependency, I'd prefer just using the bash script and not add a dev-dependency.

_If I am being too finicky about this, please go ahead and merge it_.

.travis.yml
```diff
@@ -1,13 +1,14 @@
...
+  include:
+    - node_js: "4"
+    - node_js: "6"
+    - node_js: "8"
+      after_success: "npm run report-coverage"
```
sudo-suhas: Since we use yarn on CI, change to `yarn report-coverage`?

davidtheclark:
> See https://docs.codecov.io/docs/about-the-codecov-bash-uploader.

Ah, thanks! Works like a charm.

sudo-suhas:
My pleasure @davidtheclark!

PR #147, opened 2018-05-21
fix: Support loading JS config from relative paths

evocateur:
A tool based on [kcd-scripts](https://github.com/kentcdodds/kcd-scripts) that passes relative paths to `cosmiconfig().load()` (via `lint-staged --config ./node_modules/kcd-scripts/dist/config/lintstagedrc.js`) broke after lint-staged [upgraded to cosmiconfig 5](https://github.com/okonet/lint-staged/pull/444), and I believe I've narrowed it down to a missing `path.resolve()`.

There is a functioning workaround, `require.resolve()` the argument before passing it, but this is inconsistent with many other CLIs that accept `--config` parameters, such as [eslint](https://eslint.org/docs/user-guide/command-line-interface#-c---config), rollup (lol rollupjs.org is borked right now), and [prettier](https://prettier.io/docs/en/cli.html#find-config-path-and-config):
```
eslint --config ./dist/config/eslint.js
rollup --config ./dist/config/rollup.js
prettier --config ./dist/config/prettier.js
```

This pattern worked with cosmiconfig 4.

### DIFF
--- a/test/successful-directories.test.js
+++ b/test/successful-directories.test.js
@@ -1027,20 +1027,20 @@ describe('works fine if sync loader returns a Promise from a JS file', () => {
   const startDir = temp.absolutePath('a/b/c/d/e/f');
   beforeEach(() => {
     temp.createFile(
-      'a/b/c/d/e/f/foo.config.js',
+      'a/b/c/d/e/f/bar.config.js',
       'module.exports = Promise.resolve({ a: 1 });'
     );
   });
 
   const explorerOptions = {
     stopDir: temp.absolutePath('.'),
-    searchPlaces: ['foo.config.js'],
+    searchPlaces: ['bar.config.js'],
   };
 
   test('sync', () => {
-    const result = cosmiconfig('foo', explorerOptions).searchSync(startDir);
+    const result = cosmiconfig('bar', explorerOptions).searchSync(startDir);
     expect(result).toEqual({
-      filepath: temp.absolutePath('a/b/c/d/e/f/foo.config.js'),
+      filepath: temp.absolutePath('a/b/c/d/e/f/bar.config.js'),
       config: expect.any(Promise),
     });
     return result.config.then(resolved => {
--- a/src/createExplorer.js
+++ b/src/createExplorer.js
@@ -273,7 +273,7 @@ class Explorer {
       return cacheWrapper(this.loadCache, absoluteFilePath, () => {
         return readFile(absoluteFilePath, { throwNotFound: true })
           .then(content => {
-            return this.createCosmiconfigResult(filepath, content);
+            return this.createCosmiconfigResult(absoluteFilePath, content);
           })
           .then(this.config.transform);
       });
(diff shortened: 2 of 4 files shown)


test/successful-directories.test.js
```diff
@@ -1027,20 +1027,20 @@ describe('works fine if sync loader returns a Promise from a JS file', () => {
   const startDir = temp.absolutePath('a/b/c/d/e/f');
   beforeEach(() => {
     temp.createFile(
-      'a/b/c/d/e/f/foo.config.js',
+      'a/b/c/d/e/f/bar.config.js',
```
evocateur: These tests kept failing for me, semi-randomly, and this was quicker than figuring out what module cache issues were to blame...

chrisblossom:
Isn't this already being done here: [``src/createExplorer.js#L272``](https://github.com/davidtheclark/cosmiconfig/blob/951da3046490fa823627ea6388e2b6cd5baa5aac/src/createExplorer.js#L272)?

evocateur:
@chrisblossom To load the _content_, sure, but the JS loader ignores that argument, using the raw `filepath`: https://github.com/davidtheclark/cosmiconfig/blob/951da3046490fa823627ea6388e2b6cd5baa5aac/src/loaders.js#L7-L8

evocateur:
If I had to guess, `cosmiconfig@4` was exploiting a side-effect of `require-from-string` to successfully load this kind of content (`.js`). Now that it's just a raw `require()`, that absolute path resolution inside `load()`/`loadSync()` is essentially wasted in the `.js` case.

chrisblossom:
I thought the `absoluteFilePath` was being passed to the loader, but looks like I was wrong. I think if 
https://github.com/davidtheclark/cosmiconfig/blob/951da3046490fa823627ea6388e2b6cd5baa5aac/src/createExplorer.js#L276

is changed to use `absoluteFilePath` instead of `filePath`, this would be better solved. Unless I am missing where the loader is being called.

davidtheclark:
Thanks, @evocateur!

> If I had to guess, cosmiconfig@4 was exploiting a side-effect of require-from-string to successfully load this kind of content (.js).

Quite possible, yeah. Thanks a lot for adding the new test, @evocateur, to prevent regressions!

> changed to use absoluteFilePath instead of filePath

Yeah, that seems like the right thing to do, @chrisblossom 👍

evocateur:
Ok, changes applied. Thanks for looking at this so quickly!

test/successful-directories.test.js
evocateur: This was, as I explained, necessary to make the test suite pass at all. The snapshots kept breaking, returning an object instead of a Promise, exactly like a cached search would return the eventually-resolved config object for a duplicate path.

This test isn’t testing the cache, so this is an appropriate change, in my opinion.

davidtheclark: @chrisblossom I opened a follow-up issue about this: https://github.com/davidtheclark/cosmiconfig/issues/148

davidtheclark:
Published in 5.0.5.

evocateur:
Great, thanks!

PR #153, opened 2018-08-09
fix: `no loader` error for unsupported extension

ikatyang:
Context: prettier/prettier#4962

Fixes `Cannot read property 'sync' of undefined` for unsupported extension.


davidtheclark:
Great fix, thanks. I'll cut a release soon.

PR #157, opened 2018-11-03
clear matching require.cache entries

olsonpm:
@davidtheclark, @sudo-suhas

Jest supplies its own `require` which makes testing an issue.  I would prefer adding a test outside of jest for this case, especially since we lose out on code coverage ([jest doesn't implement `require.cache`](https://github.com/facebook/jest/issues/5120)).  If we want to stay within Jest, then I'll need to spend some time figuring out why [this block](https://github.com/olsonpm/cosmiconfig/blob/fix-require-cache/test/caches.test.js#L609-L612) fails.   `jest.resetModules` seems to be doing something asynchronous under the hood, but it will take me more time to understand what.

Please let me know what you think

### DIFF
--- a/src/createExplorer.js
+++ b/src/createExplorer.js
@@ -33,19 +33,19 @@ class Explorer {
 
   clearLoadCache() {
     if (this.loadCache) {
-      this.loadCache.clear();
+      clearCache(this.loadCache);
     }
     if (this.loadSyncCache) {
-      this.loadSyncCache.clear();
+      clearCache(this.loadSyncCache);
     }
   }
 
   clearSearchCache() {
     if (this.searchCache) {
-      this.searchCache.clear();
+      clearCache(this.searchCache);
     }
     if (this.searchSyncCache) {
-      this.searchSyncCache.clear();
+      clearCache(this.searchSyncCache);
     }
   }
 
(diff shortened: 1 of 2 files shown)


davidtheclark:
Awesome work investigating the potential pitfalls and finding the prior art, @olsonpm. Two thoughts upfront:

- I wonder if we want to clear the `require` cache immediately before using `require` [in the JS loader](https://github.com/davidtheclark/cosmiconfig/blob/c6f0953b808a9046a23f6e8983e6cbb455dd03c7/src/loaders.js#L8). That way if a user *turns off* the cache, via options, they won't also have to call `clearCache` to get this effect. The cosmiconfig cache would still be in place by default, preventing unnecessary re-`require`s of JS files.
- Why not use [import-fresh](https://github.com/sindresorhus/import-fresh) directly? (This would also ease Jest testing, because if we mock import-fresh and verify that it's called with expected parameters, we won't have to do anything fancy and strange.)

sudo-suhas:
> Why not use import-fresh directly?

This seems like a good idea to me as well.

olsonpm:
Alright [the new PR](https://github.com/davidtheclark/cosmiconfig/pull/158) should be ready for another look.

I still think that as long as we are meddling with `require.cache` there's a good chance we're breaking someone else's expectations simply because it's global state.  We can always address it if it becomes a problem though, an alternative solution isn't worth the effort yet.

davidtheclark:
> I still think that as long as we are meddling with require.cache there's a good chance we're breaking someone else's expectations simply because it's global state.

Legitimate concern. I don't know of another solution, though, and I also am having trouble thinking of a concrete scenario where this require-cache-clearing would break someone's code. Would you feel better about releasing this as a beta to see how it works, or release with a major version bump?

olsonpm:
Yeah I can't think of any use-cases, it would have to be code which relies on very specific caching behavior.  I don't think it will bubble up from a beta preview, and I assume the benefit of pushing a patch fix outweighs the cost of pushing users to a new major revision.

Reading back, it seems the decision to remove `require-from-string` was to enable users to hook into `require` correct?  If the global cache proves problematic, maybe we can bring `require-from-string` back in and expose a `require` js loader should users prefer it.

PR #160, opened 2018-11-17
cosmiconfig.cosmiconfig = cosmiconfig;

### LINKED ISSUE #159: Alternative to 'import * as' module syntax
This is no longer valid JavaScript:

```javascript
import * as cosmiconfig from 'cosmiconfig';

cosmiconfig('module-name');
```

https://stackoverflow.com/questions/39415661/what-does-resolves-to-a-non-module-entity-and-cannot-be-imported-using-this/39415662#39415662

> An ES6 module namespace object cannot be invoked as a function or with new.
> The thing you import using * as X from a module is defined to only have properties.

As a backwards-compatible change, this module should have something like this added at the bottom of `src/index.js`:

```javascript
cosmiconfig.cosmiconfig = cosmiconfig;
```

That way we'll be able to do this:

```javascript

cosmiconfig('module-name');
```

It's a bit silly. A `default` export is doable too, but as an external TypeScript user, I find it inconvenient to have a separate import syntax and/or group for importing this one function verses any associated types the type definitions also export.

JoshuaKGoldberg:
🙌 JavaScript!

Fixes #159.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -48,6 +48,7 @@ function cosmiconfig(
   return createExplorer(normalizedOptions);
 }
 
+cosmiconfig.cosmiconfig = cosmiconfig;
 cosmiconfig.loadJs = loaders.loadJs;
 cosmiconfig.loadJson = loaders.loadJson;
 cosmiconfig.loadYaml = loaders.loadYaml;


davidtheclark:
Thanks for the suggestion, but I'm not interested right now in supporting specific TypeScript usage patterns, when alternatives already work ok.

PR #169, opened 2019-01-27
Nested packageProp path

### LINKED ISSUE #164: Nested packageProp path
Using `cosmiconfig` in a yarn version tool, [`yvm`](https://github.com/tophat/yvm). It would be helpful to be able to use the yarn version in the package.json engines prop as a search place for configuration. e.g.
```json
packageProp: 'engines.yarn',
```
A custom loader would probably work, but it seemed like a useful native feature regardless.

iamogbz:
- Allows targeting of nested `package.json` values using dot notation

Closes #164

### DIFF
--- a/src/createExplorer.js
+++ b/src/createExplorer.js
@@ -2,6 +2,7 @@
 'use strict';
 
 const path = require('path');
+const get = require('lodash.get');
 const loaders = require('./loaders');
 const readFile = require('./readFile');
 const cacheWrapper = require('./cacheWrapper');
@@ -173,7 +174,7 @@ class Explorer {
 
   loadPackageProp(filepath: string, content: string) {
     const parsedContent = loaders.loadJson(filepath, content);
-    const packagePropValue = parsedContent[this.config.packageProp];
+    const packagePropValue = get(parsedContent, this.config.packageProp);
     return packagePropValue || null;
   }
 
--- a/README.md
+++ b/README.md
@@ -410,6 +410,26 @@ Default: `` `${moduleName}` ``.
 
 Name of the property in `package.json` to look for.
 
+Use periods to access nested properties.
+
+Example:
+
+```js
+{
+  packageProp: 'engines.node'
+}
+```
+
+to access the value specified in the `package.json`
+
+```json
+{
+  "engines": {
+    "node": ">=4"
+  }
+}
+```
+
 ### stopDir
 
 Type: `string`.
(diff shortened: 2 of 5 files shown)


src/createExplorer.js
```diff
@@ -173,7 +174,7 @@ class Explorer {
 
   loadPackageProp(filepath: string, content: string) {
     const parsedContent = loaders.loadJson(filepath, content);
-    const packagePropValue = parsedContent[this.config.packageProp];
+    const packagePropValue = get(parsedContent, this.config.packageProp);
```
iamogbz: Not sure where to add the tests for the nested prop logic

davidtheclark: Tests will need to be added in all the places that `package.json` parsing is tested.

iamogbz: @davidtheclark done!

iamogbz:
@davidtheclark is there anything else needed for this?

src/createExplorer.js
olsonpm: We're taking a risk here that current devs aren't using a period in `packageProp`, which might be fine.  Just wanted to point out the possible breaking change.

olsonpm:
@iamogbz - please update [the readme](https://github.com/davidtheclark/cosmiconfig/blob/master/README.md#packageprop) with a small blurb on allowing nested package properties

other than that it looks good to me.  At first I was weary of pulling in lodash.get because lodash utils are bigger than most people think - but its functionality is ubiquitous and thus expected, and the size isn't _that_ big.

iamogbz:
> @iamogbz - please update [the readme](https://github.com/davidtheclark/cosmiconfig/blob/master/README.md#packageprop) with a small blurb on allowing nested package properties

@olsonpm updated, thanks!

src/createExplorer.js
iamogbz: True, I guess we could use

```js
parsedContent[this.config.packageProp] || get(parsedContent, this.config.packageProp);
```

to mitigate the breaking change?

olsonpm: ha yeah I guess that's easy enough.  :thumbsup:

iamogbz: So it seems `lodash.get` already handles this case 😄 

Check it out https://runkit.com/iamogbz/5c60e0e12da4920012e32cad

olsonpm: oye - til I learned.  Thanks much for the quick code snippet.

davidtheclark:
Thanks @iamogbz. It's looking good. I think there may be a bunch of additional tests worth writing for cases other than successfully loading a named file. Do you mind if I add a commit to this PR?

Issue #172, opened 2019-02-22
Option to disable package.json special-casing


cspotcode:
In rare cases, I want to use cosmiconfig in a generic way, to load from a user-specified config file.  I want my code to behave identically regardless of the filename specified by the user, so I need to disable cosmiconfig's special-casing of `package.json`.

One hack is to ask for `./package.json` instead of `package.json`.  However, I think a better approach is for `Explorer` to have a property `packageJsonName` which is a string or array of strings listing filenames that should be treated as `package.json` instead of a normal config file.

This property can be modified by people like me who need to treat package.json like any other config file.

davidtheclark:
@cspotcode `package.json` is a very loaded file name. You really want it to be treated like any other configuration file? I don't think I understand your use case. Could you please explain more?

cspotcode:
There are 2 possible use-cases:

a. Treating another file, perhaps `manifest.json`, the same way as `package.json`.  This might be useful when developing Chrome extensions.

b. Using cosmiconfig to discover the location of a config file without actually loading it.

I started investigating cosmiconfig as a replacement for findup-sync in Mocha's new v6.  findup-sync is a big contributor to slow mocha startup time, exacerbated by the fact that mocha needs to spawn an extra process internally.  (the process spawning is necessary for good reasons)  cosmiconfig loads much faster.

However, I had to disable cosmiconfig's config loading, because Mocha already has specific ways it loads config files.  Setting `loaders: '.js': noopFn, '.json': noopFn, //...` does the trick.  Unfortunately, there's no override for how `package.json` files are loaded.  And since I need mocha to use its existing logic to avoid breaking changes and preserve its diagnostic messages, I need to disable cosmiconfig's special-casing behavior.  (I did this with `packageProp: '__proto__'` but that's really hacky)

An alternative is exposing an API to locate files but not load them.  This could be an iterator, compatible with for-of.  Keep calling it and it keeps returning the next candidate config file.

---

I realize my use-case might not be very common, but the implementation in #173 is very clean.  Instead of a magic, hardcoded string, it's a config value.  All the logic is still just as simple, and default behavior is identical to today.

Your README says the search process is highly customizable; this change is in the same spirit.

davidtheclark:
> b. Using cosmiconfig to discover the location of a config file without actually loading it.

This seems like a nice feature on its own, independent of any `package.json`-specific use cases.

> there's no override for how `package.json` files are loaded

[...]

PR #173, opened 2019-02-22
Implement packageFilenames: specify which files get the package.json special treatment

cspotcode:
Implements #172.

Adds a new option `packageFilenames: null | string | Array<string>;`  Defaults to `['package.json']`.  Controls which config files will be given special treatment as a `package.json` file.

If you want to pull config from a `packageProp` of an alternative file, say `manifest.json`, you can do that.
You can also disable special-casing entirely, so that `package.json` can be loaded through the same code paths as any other config file.

### DIFF
--- a/flow-typed/defs.js
+++ b/flow-typed/defs.js
@@ -21,7 +21,7 @@ type ExplorerOptions = {
   ignoreEmptySearchPlaces: boolean,
 };
 
-type ExplorerContext =   & {
+type ExplorerContext = ExplorerOptions & {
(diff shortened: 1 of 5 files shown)


cspotcode:
Updated; tests are passing; coverage is 100%.

cspotcode:
I ended up using find-up in mocha, so my use-case for this feature doesn't exist anymore.  I appreciate everyone taking the time to discuss.  mochajs/mocha#3830

PR #174, opened 2019-02-28
Migrate to the lodash package

stof:
The modularized lodash packages are unmaintained since 2 years, lagging a lot behind the main lodash package. Partial imports of lodash are the preferred way.
See https://github.com/lodash/lodash/issues/2969

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "import-fresh": "^2.0.0",
     "is-directory": "^0.3.1",
     "js-yaml": "^3.9.0",
-    "lodash.get": "^4.4.2",
+    "lodash": "^4.17.11",
     "parse-json": "^4.0.0"
   },
   "devDependencies": {
--- a/src/createExplorer.js
+++ b/src/createExplorer.js
@@ -2,7 +2,7 @@
 'use strict';
 
 const path = require('path');
-const get = require('lodash.get');
+const get = require('lodash/get');
 const loaders = require('./loaders');
 const readFile = require('./readFile');
 const cacheWrapper = require('./cacheWrapper');
--- a/yarn.lock
+++ b/yarn.lock
@@ -2647,14 +2647,14 @@ locate-path@^2.0.0:
     p-locate "^2.0.0"
     path-exists "^3.0.0"
 
-lodash.get@^4.4.2:
-  version "4.4.2"
-  resolved "https://registry.yarnpkg.com/lodash.get/-/lodash.get-4.4.2.tgz#2d177f652fa31e939b4438d5341499dfa3825e99"
-  integrity sha1-LRd/ZS+jHpObRDjVNBSZ36OCXpk=
-
 lodash@^4.14.0, lodash@^4.15.0, lodash@^4.17.4, lodash@^4.2.0, lodash@^4.3.0:
   version "4.17.5"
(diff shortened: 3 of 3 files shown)


olsonpm:
personally i'd prefer the unmaintained package as it both works and doesn't require all of lodash to be npm installed.  Either that or we should remove the dependency altogether and write the `_.get` utility ourselves.

chrisblossom:
I'd think a maintained version should be used. What about replacing `lodash.get` with [get-value](https://www.npmjs.com/package/get-value)?

stof:
I created #175 using `get-value`

cspotcode:
Startup performance should be prioritized over download size, right?  Cosmiconfig is one of the first libraries loaded by CLI tools.  It should load as fast as possible.  It gets installed only once, along with lots of other things.  However, it is loaded hundreds of times, and that's where developers are going to notice any slowdown.

I was curious so I did a benchmark.  Scroll to the bottom for results. https://github.com/cspotcode/repros/tree/cosmiconfig-lodash-perf

Load times are impacted by fs calls.  `require('lodash')` pulls a single file.  `require('lodash/get')` pulls 52 files from the filesystem, and each file loaded involves multiple filesystem calls because of node's require() behavior.  This means `require('lodash.min.js')` is going to be just as fast if not faster than `require('lodash/get')`.  Even `require('get-value')` pulls more files than `require('lodash')`, and it's already a no-go because it doesn't support node v4.

Remember fs perf hit is exacerbated on Windows.  I assume there are tools using cosmiconfig that care about supporting Windows.

`require('lodash.get')` only pulls a single file and it doesn't matter if it's un-maintained.  It does its job.  Are we going to get important security patches for lodash.get?  I doubt it.

Full `lodash` is still a better option IMHO, because it's more likely to already exist in npm cache and/or in memory, loaded by another library.  So there may be *zero* download time and *zero* added load time.

olsonpm:
or we could write all 5 lines of code in our package.

cspotcode:
Yeah, exactly.  Using tons of tiny modules is kinda crazy.

Lodash.get supports other extraneous features you probably don't want, like escaping periods and `[]` syntax.  As long as you support arrays in addition to strings, I think you've covered all the bases.

```
function get(o, pathStringOrArray) {
    const path = typeof pathStringOrArray === 'string' ? pathStringOrArray.split('.') : pathStringOrArray;
    return path.reduce((a, n) => a[n], o);
}
```

davidtheclark:
I'm not sure we need to worry about the fact that `lodash.get` isn't going to receive updates, because it already does what we need it to do. We could just fix the version and leave it at that.

I also think it would be perfectly reasonable to add a new, thoroughly unit-tested function that does that same thing in this codebase — no dependency. We have a simpler situation than a generic library, because we *don't* need to support arrays (I think that would be a crazy way to use `package.json`). So let's do this.

cspotcode:
The only gotcha is if config is stored like this:

```
// package.json
{
    "build.release": {/* config goes here */}
}
```

You'll want people to be able to specify: `packageProp: ["build.release"]` to avoid ambiguity.

lodash's `get` behavior is weird.  It'll actually try multiple variations: both splitting on periods and not splitting.  And it'll return one of it finds it.  I forget the exact search order.

davidtheclark:
I'm going to create a new PR to include our own `get`-like function. Good point about the arrays, @cspotcode: I'll take that into account.

PR #175, opened 2019-03-01
Replace the unmaintained lodash.get with get-value

stof:
This replaces #174, using `get-value` rather than the full `lodash`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,10 +83,10 @@
     ]
   },
   "dependencies": {
+    "get-value": "^3.0.1",
     "import-fresh": "^2.0.0",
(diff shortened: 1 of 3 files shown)


stof:
hmm, looks like `get-value` does not support node 4. Is this version of Node still relevant ?

olsonpm:
some of the larger libraries using cosmiconfig support node v4 which is why davidtheclark didn't want to update.

davidtheclark:
Thanks @stof. I'm going to create a new PR to include our own `get`-like function (see https://github.com/davidtheclark/cosmiconfig/pull/174#issuecomment-468945111).

PR #177, opened 2019-03-02
Replace lodash.get with local function; support array paths.

davidtheclark:
@olsonpm or @sudo-suhas for review, please.

cc @stof, @cspotcode

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,17 +405,31 @@ Examples:
 
 ### packageProp
 
-Type: `string`.
+Type: `string | Array<string>`.
 Default: `` `${moduleName}` ``.
 
 Name of the property in `package.json` to look for.
 
-Use periods to describe a path to nested properties. For example, the value `'configs.myPackage'` will get you the `"myPackage"` value in a `package.json` like this:
+Use a period-delimited string or an array of strings to describe a path to nested properties.
+
+For example, the value `'configs.myPackage'` or `['configs', 'myPackage']` will get you the `"myPackage"` value in a `package.json` like this:
 
 ```json
 {
   "configs": {
     "myPackage": {..}
+  }
+}
+```
+
+If property names within the path include periods, you need to use an array of strings. For example, the value `['configs', 'foo.bar', 'baz']` will get you the `"baz"` value in a \`package.json like this:
+
+```json
+{
+  "configs": {
+    "foo.bar": {
+      "baz": {..}
+    }
   }
 }
 ```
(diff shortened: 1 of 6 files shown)


cspotcode:
Looks good to me. 👍

olsonpm:
I've got things goin on til sunday - will review then.

src/getPropertyByPath.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+'use strict';
+
+// Resolves property paths defined with period-delimited strings or arrays of strings.
+// Property names that include periods are only understand in array paths.
```
olsonpm: Had a moment to browse now - this differs from the `_.get` implementation which special cases a string to first test for the key before splitting on a period.

For backward comaptibility (even before _.get was introduced) we could do the same e.g.
```js
// getPropertyByPath.js

// Resolves property paths defined with period-delimited strings or arrays of strings.
function getPropertyByPath(source: Object, path: string | Array<string>): any {
  // first attempt the path as though it were a single key regardless whether it has periods
  if (typeof path === 'string' && source.hasOwnProperty(path))
    return source[path];

  const parsedPath = typeof path === 'string' ? path.split('.') : path;
  let config = source;
  for (const key of parsedPath) {
    if (!config || !config.hasOwnProperty(key))
      return;

    config = config[key];
  }

  return config;
}
```

and

```js
// getPropertyByPath.test.js

const source = {
  'ant.man': 'superhero',
}

test('attempts the string key before splitting', () => {
  expect(getPropertyPath(source, 'ant.man')).toBe(source['ant.man']);
});
```

davidtheclark: Good point, @olsonpm. I added a commit to address this.

README.md
```diff
@@ -420,6 +422,18 @@ Use periods to describe a path to nested properties. For example, the value `'co
 }
 ```
 
+If property names within the path include periods, you need to use an array of strings. For example, the value `['configs', 'foo.bar', 'baz']` will get you the `"baz"` value in a \`package.json like this:
```
olsonpm: this is such an edge case that I doubt it matters, but with the additional comment maybe change this to

```
If **nested** property names within the path include periods
```

davidtheclark: Yes, this is a very ugly edge case. But sure I'll make that change.

olsonpm: after that I think we're good then.  If people get confused for some reason we can amend wording later.

Thanks for coding this.

PR #179, opened 2019-03-22
Upgrade js-yaml to avoid Denial of Service

Teamop:
Based on the [npm audit](https://www.npmjs.com/advisories/788), there is an security issue with `js-yaml`, so upgrade to the latest version to fix.


davidtheclark:
Thanks @Teamop. I just merged https://github.com/davidtheclark/cosmiconfig/pull/181, which was the same change. I'll release it today.

PR #186, opened 2019-04-23
Fail CI if package.json and yarn lock are out of sync

koenpunt:
This would prevent builds for PRs that only update the `package.json` to succeed.


olsonpm:
ha - much appreciated.

davidtheclark:
Thanks!

Issue #188, opened 2019-05-31
Version 5.2.0+ -- TypeError: Cannot read property 'reduce' of undefined


jamonholmgren:
When upgrading [gluegun](https://github.com/infinitered/gluegun) from Cosmiconfig 5.1.0 to 5.2.0, I'm getting an error in my tests:

```
  ● runs a command explicitly

    TypeError: Cannot read property 'reduce' of undefined

      at getPropertyByPath (node_modules/cosmiconfig/dist/getPropertyByPath.js:15:21)
      at Explorer.loadPackageProp (node_modules/cosmiconfig/dist/createExplorer.js:177:30)
      at Explorer.loadFileContent (node_modules/cosmiconfig/dist/createExplorer.js:230:12)
      at Explorer.createCosmiconfigResultSync (node_modules/cosmiconfig/dist/createExplorer.js:263:31)
      at Explorer.loadSearchPlaceSync (node_modules/cosmiconfig/dist/createExplorer.js:158:17)
      at Explorer.searchDirectorySync (node_modules/cosmiconfig/dist/createExplorer.js:136:21)
      at run (node_modules/cosmiconfig/dist/createExplorer.js:108:27)
      at cacheWrapper (node_modules/cosmiconfig/dist/cacheWrapper.js:14:18)
      at Explorer.searchFromDirectorySync (node_modules/cosmiconfig/dist/createExplorer.js:117:14)
      at Explorer.searchSync (node_modules/cosmiconfig/dist/createExplorer.js:102:17)

  ● runs a command via passed in args

    TypeError: Cannot read property 'reduce' of undefined

      at getPropertyByPath (node_modules/cosmiconfig/dist/getPropertyByPath.js:15:21)
      at Explorer.loadPackageProp (node_modules/cosmiconfig/dist/createExplorer.js:177:30)
      at Explorer.loadFileContent (node_modules/cosmiconfig/dist/createExplorer.js:230:12)
      at Explorer.createCosmiconfigResultSync (node_modules/cosmiconfig/dist/createExplorer.js:263:31)
      at Explorer.loadSearchPlaceSync (node_modules/cosmiconfig/dist/createExplorer.js:158:17)
      at Explorer.searchDirectorySync (node_modules/cosmiconfig/dist/createExplorer.js:136:21)
      at run (node_modules/cosmiconfig/dist/createExplorer.js:108:27)
[...]

Issue #183, opened 2019-04-16
Code injection security vulnerability at js-yaml


em0ney:
Hi there,

Thanks for making cosmiconfig!

You have a dependency on `js-yaml@3.13.0`. Please see this report of a High severity vulnerability in this module.

https://www.npmjs.com/advisories/813

olsonpm:
thanks for the report.  I'll have time to take a look this weekend.

It's the second security issue we've had within a short time frame though, I'll probably check out alternative libraries while I'm at it.

em0ney:
I might open up a PR for this one if there is a chance you can get it merged this week

olsonpm:
if your issue is time sensitive then you should fork the repo and depend on that in the meantime.  I assume this issue will be addressed within the next couple weeks, but we can't promise anything.

sudo-suhas:
Removing the lock file and a clean install would pull in the latest version of `js-yaml`. So for anybody who wishes to fix this issue immediately, there is no need to wait for a fix from this library(we'll do it though).

olsonpm:
Alright, here's what I found.

js-yaml has three advisories which were raised on npm
- https://www.npmjs.com/advisories/16: published Jun 23rd, 2013
- https://www.npmjs.com/advisories/788: published Mar 21st, 2019
- https://www.npmjs.com/advisories/813: published Apr 15th, 2019

Two of them are irrelevant to cosmiconfig because the risk was
associated with [`load`](https://github.com/nodeca/js-yaml#load-string---options-).  Cosmiconfig uses [`safeLoad`](https://github.com/nodeca/js-yaml#safeload-string---options-)

However because of the existence of `load` vs `safeLoad`, and js-yaml providing the
functionality to parse js functions and regexes in `load`, I think it's safe to assume more
advisories will be raised which end up being irrelevant to Cosmiconfig.

I found two other yaml parsing libraries that seem reliable but neither support node v4

[yaml](https://www.npmjs.com/package/yaml) requires v6+ (seems more actively maintained than yaml-js)
[yaml-js](https://www.npmjs.com/package/yaml-js) requires v8+

Both libraries
- have no security advisories raised
- don't parse js functions or regexes and thus don't have a `load` vs `safeLoad` concept

Unfortunately I think we're stuck bumping js-yaml versions every time an advisory is raised for the
duration of cosmiconfig v5 which will always support node v4.  We should probably switch to
[...]

PR #190, opened 2019-06-08
Remove support for Node 4

davidtheclark:
This PR removes support for Node 4 and runs tests in Node 12 on CI.

Here are some reasons I think we may as well remove Node 4 support:

- Smooth the way going forward. If we want to be able to upgrade or shift dependencies, we are held back by Node 4 — for example, @olsonpm's [look into alternative YAML parsers](https://github.com/davidtheclark/cosmiconfig/issues/183#issuecomment-485171989).
- This library has been very stable, so any consumer that still wants to support Node 4 *can*, with versions that have already been published.
- We should publish some type checking of arguments (see https://github.com/davidtheclark/cosmiconfig/issues/188); and although I think we could do that with a patch release, it's a little nicer to do it with a major release. That opens the gate for removing support for Node 4.

@olsonpm and @sudo-suhas for review, please.

We can follow up with any changes that you think are worthwhile and unblocked by dropping Node 4 support.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,7 +14,7 @@
     "no-var": "off",
     "prefer-const": "off",
     "prefer-arrow-callback": "off",
-    "node/no-unsupported-features": ["error", {"version": 4}],
+    "node/no-unsupported-features": ["error", {"version": 6}],
     "func-names": ["error", "always"],
     "prefer-template": "error",
     "object-shorthand": [
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,7 +14,7 @@
     "no-var": "off",
     "prefer-const": "off",
     "prefer-arrow-callback": "off",
-    "node/no-unsupported-features": ["error", {"version": 6}],
+    "node/no-unsupported-features": ["error", {"version": 8}],
     "func-names": ["error", "always"],
     "prefer-template": "error",
     "object-shorthand": [
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,9 @@ branches:
     - master
 matrix:
   include:
-    - node_js: "6"
     - node_js: "8"
-      after_success: "bash <(curl -s https://codecov.io/bash)"
     - node_js: "10"
+      after_success: "bash <(curl -s https://codecov.io/bash)"
     - node_js: "12"
 install:
   - yarn install --frozen-lockfile
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,10 @@ branches:
     - master
 matrix:
   include:
-    - node_js: "4"
     - node_js: "6"
     - node_js: "8"
       after_success: "bash <(curl -s https://codecov.io/bash)"
     - node_js: "10"
+    - node_js: "12"
 install:
   - yarn install --frozen-lockfile
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## HEAD
 
-- **Breaking change:** Remove support for Node 4.
+- **Breaking change:** Remove support for Node 4 and 6. Requires Node 8+.
 
 ## 5.2.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## HEAD
+
+- **Breaking change:** Remove support for Node 4.
+
 ## 5.2.1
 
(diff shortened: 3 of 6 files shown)


olsonpm:
this is awesome.  i'm just starting a new job but would like to put some thought into this PR so please give me a couple weeks to find some time.

davidtheclark:
@olsonpm Sure 👍 I don't think we need to put all the work you might have in mind into this PR, though. This can simply remove support; then we can follow up with refactoring or dependency adjustments.

sudo-suhas:
@davidtheclark I agree with you. However, node.js 6 has also reached [end-of-life](https://github.com/nodejs/Release#end-of-life-releases) as well. Is there any reason to continue supporting node 6?

davidtheclark:
@sudo-suhas We have some reasons to remove support for Node 4. Do we have any reasons *to* remove support for Node 6?

It's true that Node 6 is EOL, but it entered that state very, very recently (4/30), so I'm not inclined to remove support, especially unless we have good reasons to do so.

olsonpm:
dependencies moving forward is the only reason i can think of.  I don't have an opinion either way though.

sudo-suhas:
> I'm not inclined to remove support, especially unless we have good reasons to do so.

@davidtheclark That makes sense, no objections from me. I suggested removing support because:

- The rationale for removing node 4 support kinda applies to node 6 as well:
  > This library has been very stable, so any consumer that still wants to support Node 4 _can_, with versions that have already been published.
- Since node 6 is now deprecated, it is reasonable to assume that other libraries would also drop support for it. Although I don't foresee this as a problem for us, dropping node 6 would be a sure shot way to ensure we don't.

chrisblossom:
If you are going to do a major semver and drop node 4, I think you should also remove support for node 6. As @sudo-suhas pointed out it is end of life and I think libraries holding on to old versions of node are doing the community a disservice by not promoting moving forward. Also, as @olsonpm said, dependencies have already started dropping node 6 support.

Already requires node 8:
[sindresorhus/parse-json](https://github.com/sindresorhus/parse-json) _**(dependency)**_
[sindresorhus/del](https://github.com/sindresorhus/del)
[eslint-plugin-node](https://github.com/mysticatea/eslint-plugin-node)
[eslint-plugin-flowtype](https://github.com/gajus/eslint-plugin-flowtype)
[husky](https://github.com/typicode/husky)
[lint-staged](https://github.com/okonet/lint-staged)
[sindresorhus/make-dir](https://github.com/sindresorhus/make-dir)
[sindresorhus/parent-module](https://github.com/sindresorhus/parent-module)


Soon to drop support:
[sindresorhus/import-fresh](https://github.com/sindresorhus/import-fresh) _**(dependency)**_
[jest 25](https://github.com/facebook/jest/pull/8455)
[eslint v6](https://github.com/eslint/eslint/blob/v6.0.0-alpha.0/CHANGELOG.md)

I think a good idea would be to support node 4 and 6 on the current version with bug fixes only.

olsonpm:
regardless of whether going forward with v6 or v8 support is the right choice, "doing the community a disservice" is a little strong.  Bumping node versions takes time, especially for larger projects.  There are of pros and cons for pushing the supported version forward, and one major drawback is forcing open source devs into volunteering more time to bumping the node version.

chrisblossom:
> regardless of whether going forward with v6 or v8 support is the right choice, "doing the community a disservice" is a little strong

Sorry I wasn't trying to come off as attacking, just an opinion of mine. I actually thought the same thing as @davidtheclark not long ago when a lot of my dependencies on my projects were dropping node 6 support (mainly [sindresorhus's](https://github.com/sindresorhus) modules). But I thought about it more and think it is actually a good thing to support the community to move forward by giving it a little push where you can.

Crazy to think that [node 6 was released in 2016](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V6.md) (seems like yesterday)...

>  one major drawback is forcing open source devs into volunteering more time to bump the node version.

The opposite is true as well, forcing open source devs to support an old version of node.

davidtheclark:
@chrisblossom's dependency list helps a lot. Ok, I feel fine about removing support for Node 6 — as noted above: this is stable; we could release bug fixes for older versions if pressured to; we haven't had bug reports so not too worried about that.

I'll get around to this within the next couple of days, I think.

davidtheclark:
Ok, Node 6 support is removed, also, for the next release.

PR #193, opened 2019-06-13
use prettier for all supported file types (except markdown)

chrisblossom:
~~removes remark dependency as well~~

Uses `remark-preset-davidtheclark` to format markdown.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "lint:md": "npm run lint:md-partial -- *.md",
     "lint:fix": "eslint . --fix",
     "lint": "eslint . && npm run lint:md",
-    "format": "prettier --write \"{src/*.js,test/*.js}\"",
+    "format": "prettier \"**/*.{js,json,yml,yaml}\" --write",
     "pretest": "npm run lint && flow check",
     "test": "jest --coverage",
     "test:watch": "jest --watch",
@@ -27,6 +27,10 @@
   "lint-staged": {
     "*.js": [
       "eslint --fix",
+      "prettier --write",
+      "git add"
+    ],
+    "*.{json,yml,yaml}": [
       "prettier --write",
       "git add"
     ],
--- a/package.json
+++ b/package.json
@@ -7,8 +7,7 @@
     "dist"
   ],
   "scripts": {
-    "lint:md-partial": "remark -u remark-preset-davidtheclark --frail --quiet --no-stdout --output --",
-    "lint:md": "npm run lint:md-partial -- *.md",
+    "lint:md": "remark-preset-davidtheclark --format",
     "lint:fix": "eslint . --fix",
     "lint": "eslint . && npm run lint:md",
     "format": "prettier \"**/*.{js,json,yml,yaml}\" --write",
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -254,46 +254,34 @@ Examples, with a module named `porgy`:
...
-  '.porgyrc',
-  'porgy.config.js'
-]
-
-// ESLint searches for configuration in these places:
-[
-  '.eslintrc.js',
+['package.json', '.porgyrc', 'porgy.config.js'][
+  // ESLint searches for configuration in these places:
+  ('.eslintrc.js',
```
davidtheclark: This formatting doesn't seem right.

package.json
```diff
@@ -8,11 +8,9 @@
   ],
   "scripts": {
     "precommit": "lint-staged && jest && flow check",
-    "lint:md-partial": "remark -u remark-preset-davidtheclark --frail --quiet --no-stdout --output --",
```
davidtheclark: Is Prettier going to automatically add the table of contents to the README? It's not, right?

I'd rather stick with remark for that reason (and I prefer the formatting over the changes above). We could use the CLI in remark-preset-davidtheclark directly, to remove the explicit remark-cli dependency, if we wanted to.

package.json
```diff
@@ -8,11 +8,9 @@
...
-    "lint:md": "npm run lint:md-partial -- *.md",
     "lint:fix": "eslint . --fix",
-    "lint": "eslint . && npm run lint:md",
-    "format": "prettier --write \"{src/*.js,test/*.js}\"",
+    "lint": "eslint .",
+    "format": "prettier \"**/*.{js,mjs,ts,json,md,yml,yaml}\" --write",
```
davidtheclark: Let's just include the file formats that we actually have in this list.

chrisblossom: I updated it to `"format": "prettier \"**/*.{js,json,md,yml,yaml}\" --write",`, I that what you mean?

README.md
chrisblossom: Prettier will format code inside markdown code blocks if it supports the language. If you end up wanting prettier on the readme / other markdown files, I'll want to update some of the examples so it is better understood as actual javascript.

package.json
chrisblossom: Correct, it does not generate the table of contents (I didn't know remark did either).

Reason I removed remark was because I thought it was unneeded since prettier can handle the formatting instead and is already used.

Should I add remark back and remove markdown files from prettier in this PR or fix the javascript formatting issues mentioned above?

chrisblossom: Markdown formatting has been removed from the PR

README.md
chrisblossom: Markdown has been removed from the PR

chrisblossom:
I copied how you are using your `remark-preset-davidtheclark` cli. Not sure if I should include the format option?

package.json
```diff
@@ -30,8 +29,12 @@
       "prettier --write",
       "git add"
     ],
+    "*.{json,yml,yaml}": [
+      "prettier --write",
+      "git add"
+    ],
     "*.md": [
-      "npm run lint:md-partial",
+      "npm run lint:md",
```
davidtheclark:
```suggestion
      "remark-preset-davidtheclark --format",
```

I think this should do it.

chrisblossom: Updated the lint-staged and npm script to include the `--format` option

PR #195, opened 2019-06-13
Node 8 updates

chrisblossom:
With node 8 you can enable trailing commas for everything. Also, eslint should target node 8.9 because that is the first LTS release and no additional features were added after, such as object spread.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
   },
   "homepage": "https://github.com/davidtheclark/cosmiconfig#readme",
   "prettier": {
-    "trailingComma": "es5",
+    "trailingComma": "all",
     "singleQuote": true,
     "printWidth": 80,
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -96,7 +96,7 @@
     "eslint-config-davidtheclark-node": "^0.2.2",
     "eslint-config-prettier": "^2.9.0",
     "eslint-plugin-flowtype": "^2.39.1",
-    "eslint-plugin-node": "^5.2.1",
+    "eslint-plugin-node": "9.1.0",
```
davidtheclark: Any reason not to use a caret here, like the other `devDependencies`?

chrisblossom: Nope, just a mistake. I rebased to master and this isn't needed anymore

Issue #191, opened 2019-06-12
[Discussion] Potential breaking changes for v6


chrisblossom:
Looks like #190 will be released soon and will require a major version bump.

Are there any potential changes to the API that should be considered before releasing `6.0.0`?

I personally would like to see this repository converted to typescript as it has a much larger community backing. I've recently made the switch and with some exceptions, I have found it a much more pleasant development experience. I'd be willing to put together an initial PR if that is something @davidtheclark @sudo-suhas @olsonpm is interested in. If not I completely understand...it isn't needed and somewhat "bikeshedding".

davidtheclark:
Thanks for opening this, @chrisblossom. 

Re TypeScript: At this point, I wouldn't mind switching to TypeScript — it would be a good educational opportunity for me, since I don't have experience with it — especially if you're interested in putting in the legwork. I'm assuming it won't require significant refactoring of non-type code: if it does, I'd be more hesitant. If you want to start a PR towards that, go for it. If you're finding yourself making more than type-level changes though, it may be a good idea to pause and discuss those before you put in too much work.

chrisblossom:
> I'm assuming it won't require significant refactoring of non-type code: if it does, I'd be more hesitant. ... If you're finding yourself making more than type-level changes though, it may be a good idea to pause and discuss those before you put in too much work.

Should be very straight forward with minimal changes necessary. I have a fair amount of experience what it takes to convert and update the build process. FYI, I don't think I'll have time to start on it until next week at the earliest.

I think the biggest change that you'll find is common js / module interop (which I think is currently an issue, just not very known). I think the best / most future proof solution is to remove default exports and use a named export instead. Relevant discussion: https://github.com/johnagan/clean-webpack-plugin/issues/114

```js
// es modules
import { cosmiconfig } from 'cosmiconfig';

// common js
const { cosmiconfig } = require('cosmiconfig');
```

sudo-suhas:
While migrating to typescript(I too support this), we should try to use the [`--strict` compiler option](https://www.typescriptlang.org/docs/handbook/compiler-options.html). However, I think we should minimise changes required for users to upgrade `cosmiconfig` unless there is a very good reason to do so.

chrisblossom:
I put in PRs #192 #193 #194 #195 in preparation for typescript. The only one that is necessary is #194, the others are just things I noticed while updating the packages. Please close if they aren't wanted changes.

chrisblossom:
[...]

PR #197, opened 2019-06-21
Typescript Migration

chrisblossom:
This is not 100% done. There are several `// TODO comments` that need looking into.

> I'm assuming it won't require significant refactoring of non-type code: if it does, I'd be more hesitant.

If you want me to rebase this PR without the large [`refactor to use async await commit`](https://github.com/davidtheclark/cosmiconfig/commit/8cc41a02cadf1fe107fc89dbc8c09155e0c85fed) and put a PR in for it afterwards let me know. I personally think this makes the code much more readable. I also tried to make `async` and `sync` versions of functions look as similar as possible.

Most of the changes I made in the [`flow —> typescript commit`](https://github.com/davidtheclark/cosmiconfig/commit/617bdbd976bfeabecffc91aa492575d8aa5623e3) were to either directly convert to typescript or to provide more accurate type coverage (changes that would have been needed with flow as well).

I thought the [`move and delete unnecessary files commit`](https://github.com/davidtheclark/cosmiconfig/commit/d7dfac293311dfaf8e141de97928fc29d19e6051) would help preserve github history, but github lacks with git renaming files. From [what I've researched](https://www.google.com/search?newwindow=1&client=firefox-b-1-d&channel=cus&ei=oyUMXaSSIKjE0PEPtpO84Ao&q=git+file+rename+history&oq=git+file+rename+history&gs_l=psy-ab.3..0j0i8i30.27554.28197..30825...0.0..0.134.596.1j4......0....1..gws-wiz.......0i71j35i39j0i7i30j0i7i10i30j0i8i7i30.yyoIZHaJ0xc), there isn't a way to fix github here. You can see the changes locally a lot easier using [sourcetree](https://www.sourcetreeapp.com/) (or equivalent) or command line.

[`replace is-directory with path-type commit`](https://github.com/davidtheclark/cosmiconfig/commit/eb34e6b78846e555703696a2ec775eeb714e72b4) put as a separate commit in case anyone had any objections to this change. I added tests to verify the same issue `is-directory` resolved is still fixed.

#191


olsonpm:
Thanks for the time you put into this :)

src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+import * as loaders from './loaders';
+
+function identity<T>(x: T): T {
+  return x;
+}
+
+// TODO: can it be undefined to disable a loader?
+interface RawLoaders {
+  [key: string]: LoaderEntry | SyncLoader | AsyncLoader;
+}
```
chrisblossom: Is it expected for a user to be able to remove a file type loader by providing an undefined/null value value? If so I'll need to add `void` here. (I'm not requesting this, just want to have the types be correct)

src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+  const result = Object.keys(rawLoaders).reduce(
+    (result: Loaders, ext: string): Loaders => {
+      const entry = rawLoaders && rawLoaders[ext];
+
+      if (typeof entry === 'function') {
+        // TODO: sync loader can be async here
+        // TODO: direct remove function support and specify both async and sync?
+        const sync: SyncLoader = entry;
+        const async: AsyncLoader = entry;
+
```
chrisblossom: Currently a user can provide an async function to to `sync`, which will return unexpected result. Also, as far as I know, the loaders option cannot be properly typed. I think we should drop the function support and only support the `{ sync: () => 'syncLoader', async: async () => 'asyncLoader' }` option object.

src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+  return result;
+}
+
+// TODO: this can be async but no way to type it and have sync functions be happy.
+// TODO: split sync/async options like loaders?
+export type Transform = (
+  CosmiconfigResult: CosmiconfigResult,
+) => CosmiconfigResult;
+// | Promise<CosmiconfigResult>;
+
```
chrisblossom: I think there should be a way to have a `sync` and `async` transform. Also, there isn't a way to properly type the function at its current state.

Options: 
- Use options object `{ sync: () => 'transformSync', async: async () => 'transformAsync' }`.
- Add `transformSync` option.
- Deprecate the direct `transform` option and add: `transformSync` / `transformAsync` options.
- No API changes. Uncomment out `Promise<CosmiconfigResult>` and use `// @ts-ignore` comments to suppress any typescript errors. (not my preference)

src/types.ts
```diff
@@ -0,0 +1,46 @@
...
+// type ConfigPromise = Promise<ResolvedConfig>;
+
+// https://github.com/davidtheclark/cosmiconfig/issues/134
+export type Config = any;
+// | ResolvedConfig
+// | ConfigFn
+// | ConfigPromise
+// | ConfigFnPromise;
+/* eslint-enable */
+
```
chrisblossom: Do we want to try and type what we expect the config to be? I don't think we we do but I wanted to check. I think `any` is best used here.

src/types.ts
```diff
@@ -0,0 +1,46 @@
...
+// TODO: define what a config is ?
+// interface ResolvedConfig {
+//   [key: string]: unknown;
+// }
+// type ConfigFn = (...params: any[]) => ResolvedConfig;
+// type ConfigFnPromise = (...params: any[]) => Promise<ResolvedConfig>;
+// type ConfigPromise = Promise<ResolvedConfig>;
+
+// https://github.com/davidtheclark/cosmiconfig/issues/134
+export type Config = any;
```
chrisblossom: Do we want to try and type what we expect the config to be? I don't think we we do but I wanted to check. I think any is best used here.

src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+  AsyncLoader,
+} from './types';
+import { createExplorer } from './createExplorer';
+import * as loaders from './loaders';
+
+function identity<T>(x: T): T {
+  return x;
+}
+
+// TODO: can it be undefined to disable a loader?
```
chrisblossom: Is it expected for a user to be able to remove a file type loader by providing an undefined/null value value? If so I'll need to add void here. (I'm not requesting this, just want to have the types be correct)

src/index.ts
chrisblossom: Is it expected for a user to be able to remove a file type loader by providing an undefined/null value value? If so I'll need to add void here. (I'm not requesting this, just want to have the types be correct)

src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  const result = Object.keys(rawLoaders).reduce(
+    (result: Loaders, ext: string): Loaders => {
+      const entry = rawLoaders && rawLoaders[ext];
+
+      if (typeof entry === 'function') {
+        // TODO: sync loader can be async here
+        // TODO: direct remove function support and specify both async and sync?
+        const sync: SyncLoader = entry;
```
chrisblossom: Currently a user can provide an async function to to `sync`, which will return unexpected result. Also, as far as I know, the `loaders` option cannot be properly typed. I think we should drop the function support and only support the `{ sync: () => 'syncLoader', async: async () => 'asyncLoader' }` option object.

src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  return result;
+}
+
+// TODO: this can be async but no way to type it and have sync functions be happy.
+// TODO: split sync/async options like loaders?
+export type Transform = (
+  CosmiconfigResult: CosmiconfigResult,
+) => CosmiconfigResult;
+// | Promise<CosmiconfigResult>;
```
chrisblossom: I think there should be a way to have a sync and async transform. Also, there isn't a way to properly type the function at its current state.

Options:

- Use options object `{ sync: () => 'transformSync', async: async () => 'transformAsync' }`.
- Add `transformSync` option.
- Deprecate the direct `transform` option and add: `transformSync` / `transformAsync` options.
- No API changes. Uncomment out `Promise<CosmiconfigResult>` and use `// @ts-ignore` comments to suppress any typescript errors. (not my preference)

src/types.ts
```diff
@@ -0,0 +1,46 @@
...
+// | ResolvedConfig
+// | ConfigFn
+// | ConfigPromise
```
chrisblossom: Is it expected for a user to be able to remove a file type loader by providing an `undefined`/`null` value value? If so I'll need to add `void` here. (I'm not requesting this, just want to have the types be correct)

src/index.ts
davidtheclark: No, that's not documented or expected. (If someone doesn't want to use a default loader, they should adjust `searchPlaces` to avoid searching for files with that extension.)

src/index.ts
davidtheclark: I think that any API changes should be saved for followup work, even if it means weaker typing now.

src/index.ts
davidtheclark: Same as above: I think that API changes should be saved for a followup PR. If typing isn't perfect, it can be improved in the future.

chrisblossom:
> If not, are there any practices that TypeScript users have established to deal with these kinds of transitions?

I looked at Jest, which is the biggest open source project I know of that recently changed to typescript and as far as I could tell they had the same issue here. I have an idea how to work around this though.

> For a big shift like this, I think that the top priority is to make changes gradual and reviewable. So I suggest dropping conversions of Promises to async/await, or any reconsideration of APIs. We can always follow up with other changes and improvements.

I completely understand. To better minimize changes I am going to put commits and PRs in the following order:

1. Update build system to use babel and es modules.
2. Direct minimal migration from typescript -> flow (I have an idea how to get a good github diff)
3. Add code quality tools such as [`@typescript-eslint/eslint-plugin`](https://www.npmjs.com/package/@typescript-eslint/eslint-plugin) and fix any lint issues. (this was a big portion of the changed code)
4. Code PR bugs / issues found when updating types.
5. Modernize code with async await.
6. Open discussion / PRs with API change ideas.

I don't think a new version should be released until at least item number 4 is completed.

Does that sound like a better / more manageable plan?

davidtheclark:
That sounds excellent, @chrisblossom, thanks for thinking it through.

PR #198, opened 2019-06-23
es modules and build system updates

chrisblossom:
This PR tackles item number 1 in https://github.com/davidtheclark/cosmiconfig/pull/197#issuecomment-504700070.

This lays most of the groundwork to migrate to typescript, but I tried to also let this PR stand on its own.

https://github.com/davidtheclark/cosmiconfig/commit/b2b5d39b31e9247e57b0fb8954cc4feb615d0866 - Please look at this commit and the description and verify you agree with this change.
https://github.com/davidtheclark/cosmiconfig/commit/c486b8670f81f764db9c69c38c831fcaa3252d45 - This PR includes a breaking change to using a named export as [described here](https://github.com/davidtheclark/cosmiconfig/issues/191#issuecomment-501388544).

```js
// es modules

// common js
const { cosmiconfig } = require('cosmiconfig');
```


.travis.yml
```diff
@@ -9,7 +9,9 @@ matrix:
   include:
     - node_js: '8'
     - node_js: '10'
-      after_success: 'bash <(curl -s https://codecov.io/bash)'
     - node_js: '12'
 install:
   - yarn install --frozen-lockfile
+script:
+  - yarn run check:all
+  - codecov
```
davidtheclark: Why not leave this in the `after_success` hook?

appveyor.yml
```diff
@@ -26,4 +26,5 @@ install:
 test_script:
   - node --version
   - yarn --version
-  - cmd: 'yarn test'
+  - yarn run check:all
+  - codecov
```
davidtheclark: I don't think we need codecov here if it's already registered by Travis.

package.json
```diff
@@ -97,17 +108,23 @@
...
   },
   "devDependencies": {
+    "@babel/cli": "7.4.4",
+    "@babel/core": "7.4.5",
+    "@babel/plugin-proposal-class-properties": "7.4.4",
+    "@babel/preset-env": "7.4.5",
+    "@babel/preset-flow": "7.0.0",
     "babel-eslint": "^10.0.1",
-    "babel-plugin-transform-flow-strip-types": "^6.22.0",
+    "codecov": "3.5.0",
```
davidtheclark: Why does the way we deal with codecov need to change at all?

package.json
```diff
@@ -7,29 +7,29 @@
     "dist"
   ],
   "scripts": {
+    "clean": "del-cli --dot=true \"./dist/**/*\"",
+    "build": "npm run clean && npm run build:compile",
+    "build:compile": "cross-env NODE_ENV=production babel src -d dist --verbose --extensions .js --ignore **/**/*.test.js --source-maps",
```
davidtheclark: I think we should wrap this glob with globstars in quotation marks so it works across terminals, right?

package.json
```diff
@@ -7,29 +7,29 @@
...
+    "build:compile:watch": "cross-env NODE_ENV=development babel src -d dist --verbose --watch --extensions .js --ignore **/**/*.test.js --source-maps",
+    "dev": "npm run clean && npm run build:compile:watch",
```
davidtheclark: Why have watch and dev commands been added?

package.json
```diff
@@ -86,8 +86,19 @@
...
+          "targets": {
+            "node": "8.9"
+          }
+        }
+      ],
+      "@babel/preset-flow"
+    ],
     "plugins": [
-      "transform-flow-strip-types"
+      "@babel/plugin-proposal-class-properties"
```
davidtheclark: We weren't using class properties before, were we? Why do we need to add this?

README.md
```diff
@@ -70,7 +70,7 @@ Tested in Node 8+.
 Create a Cosmiconfig explorer, then either `search` for or directly `load` a configuration file.
 
 ```js
-const cosmiconfig = require('cosmiconfig');
+const { cosmiconfig } = require('cosmiconfig');
```
davidtheclark: Within the PR let's document this breaking change under the heading `HEAD` in `CHANGELOG.md`.

package.json
chrisblossom: I tried to put the why in the commit: _"codecov automatically merges tests from multiple places for the same commit"_. This includes windows and all node tests. Not necessary at all. Should I remove the commit?

package.json
chrisblossom: I've always added it as part of a standard babel setup. I'll remove it. As you said, it isn't being used.

chrisblossom:
> I think it's important to check those with every commit, so you don't need to make follow-up commits just to clean-up linting/typing errors.

This is why I wanted to point this out. I've always been taught to commit often so it is easy to go back then merge commits when finished. I just end up using the `--no-verify` arg. I'll add it back to `pre-commit`.

package.json
chrisblossom: I merged the two scripts into dev.
`"dev": "npm run clean && npm run build:compile -- --watch",`

package.json
davidtheclark: I'm not sure I understand that sentence as an explanation. Would you mind restating?

davidtheclark:
> If we keep linting and type errors, we should probably keep tests in there too, right?

I'm fine with just linting and type checking, indifferent to the tests.

package.json
chrisblossom: If code ever gets added such as:

```js
import { platform } from 'os';

if (platform() === 'win32') {
    // do windows stuff
}

if (process.version === '8') {
    // do node 8 stuff
}
```

You'll keep 100% code coverage without having to ignore the lines. This is probably unlikely though and could be dealt with if it ever comes up.

README.md
chrisblossom: Added!

davidtheclark:
> If we keep linting and type errors, we should probably keep tests in there too, right?

Tests are significantly slower and don't run only on the staged files, so I'm fine with drawing that line. There aren't too many tests here, since there's not that much code, so I also don't mind keeping the tests in the pre-commit hook. Either way.

chrisblossom:
> Tests are significantly slower and don't run only on the staged files, so I'm fine with drawing that line. There aren't too many tests here, since there's not that much code, so I also don't mind keeping the tests in the pre-commit hook. Either way.

I put the tests along with linting/type checking back into the hook.

I think the last concern (currently) I need to deal with is `codecov`. Should I remove that change or fix the windows errors?

package.json
davidtheclark: > If code ever gets added such as ... This is probably unlikely though and could be dealt with if it ever comes up.

Yeah, let's deal with it if/when it comes up.

davidtheclark:
Yeah, let's undo the codecov change until it becomes necessary; then we can merge when this has the green light.

package.json
chrisblossom: Removed

chrisblossom:
> Yeah, let's undo the codecov change until it becomes necessary; then we can merge when this has the green light.

Sounds good! Let me know if I should put PRs in a different way. 

Also, I've been git force pushing to keep the commits clean in case it gets merged. If you want me to stop doing that (and possibly do it before it is merged) let me know.

davidtheclark:
> Also, I've been git force pushing to keep the commits clean in case it gets merged. If you want me to stop doing that (and possibly do it before it is merged) let me know.

Thanks for asking. I don't have any problem with this unless I'm also contributing commits to the branch; since that's not the case here, go for it. I appreciate the effort at clean commits.

PR #202, opened 2019-07-03
Typescript code quality tools

chrisblossom:
This PR tackles item number 3 in [#197](https://github.com/davidtheclark/cosmiconfig/pull/197#issuecomment-504700070).

Again, it will probably easiest to look at each commit separately to get an idea of what was changed.

If you want me to merge the commits before merging let me know.

### DIFF
--- a/src/createExplorer.ts
+++ b/src/createExplorer.ts
@@ -10,6 +10,7 @@ import {
   LoaderEntry,
   SyncLoader,
   AsyncLoader,
+  Config,
 } from './types';
 
 const MODE_SYNC = 'sync';
@@ -18,7 +19,7 @@ const MODE_SYNC = 'sync';
 // null represents that the loader did not find anything relevant.
 // undefined represents that the loader found something relevant
 // but it was empty.
-type LoadedFileContent = Object | null | void;
+type LoadedFileContent = Config | void;
 
 class Explorer {
   private readonly loadCache: Map<string, Promise<CosmiconfigResult>> | null;
--- a/src/createExplorer.ts
+++ b/src/createExplorer.ts
@@ -76,17 +76,17 @@ class Explorer {
     });
   }
 
-  public search(searchFrom?: string): Promise<CosmiconfigResult> {
+  public async search(searchFrom?: string): Promise<CosmiconfigResult> {
     searchFrom = searchFrom || process.cwd();
-    return getDirectory(searchFrom).then((dir): Promise<CosmiconfigResult> => {
+    return getDirectory(searchFrom).then(async (dir): Promise<CosmiconfigResult> => {
       return this.searchFromDirectory(dir);
     });
   }
 
-  private searchFromDirectory(dir: string): Promise<CosmiconfigResult> {
+  private async searchFromDirectory(dir: string): Promise<CosmiconfigResult> {
     const absoluteDir = path.resolve(process.cwd(), dir);
-    const run = (): Promise<CosmiconfigResult> => {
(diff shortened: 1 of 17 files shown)


src/createExplorer.ts
```diff
@@ -255,10 +255,10 @@ class Explorer {
     content: string | null,
   ): Promise<CosmiconfigResult> {
     return Promise.resolve()
-      .then(() => {
+      .then(():Promise<LoadedFileContent> | LoadedFileContent  => {
```
davidtheclark: Should there be a space after this colon?

src/createExplorer.ts
```diff
@@ -21,6 +21,15 @@ const MODE_SYNC = 'sync';
...
 
+function nextDirUp(dir: string): string {
+  return path.dirname(dir);
+}
+
+function getExtensionDescription(filepath: string): string {
+  const ext = path.extname(filepath);
+  return ext ? `extension "${ext}"` : 'files without extensions';
+}
+
```
davidtheclark: I'd prefer not to have an ESLint rule that forces this unnecessary change: I don't think it prevents any errors, and I do think it can sometimes improve readability to take advantage of function hoisting.

src/getPropertyByPath.ts
```diff
@@ -5,7 +5,7 @@
 // understood in array paths.
 function getPropertyByPath(
   source: { [key: string]: unknown },
-  path: string | Array<string>,
+  path: string | string[],
```
davidtheclark: I prefer the generic syntax, `Array<string>`, because it stands out more clearly.

src/createExplorer.ts
chrisblossom: I ran prettier after all commits to minimize changes per commit. This comment is marked outdated so I’m going to assume prettier did it’s job. (I’m on mobile right now, I’ll check for sure when I’m on a computer.

src/getPropertyByPath.ts
chrisblossom: This comes from their recommended preset. Should I override?

src/createExplorer.ts
chrisblossom: I think you are referring to not using a function before it is defined? Personally I disagree, but don’t have strong feelings / don’t think it matters for small projects.

This comes from their recommended preset. Should I override?

olsonpm: i don't care about stylistic decisions such as this when it's a shared project, but personally i use function hoisting a lot to make my code more readable.  In my opinion it allows for the higher level code to be read first without needing to bother with the implementation of it.

src/getPropertyByPath.ts
chrisblossom: Fixed via [`'@typescript-eslint/array-type': ['error', 'generic']`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/array-type.md) - https://github.com/davidtheclark/cosmiconfig/pull/202/commits/a9e53cce3df0dff34e43eea2b18e57be302c838a

src/createExplorer.ts
chrisblossom: I was able to get back at a computer and revert this change. I also set [`'@typescript-eslint/no-use-before-define': 'off'`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-use-before-define.md).

src/createExplorer.ts
chrisblossom: Just checked, this was fixed in https://github.com/davidtheclark/cosmiconfig/pull/202/commits/19fc5ef3fd8d1bde07aee619ea00bbe2ea804429.

PR #205, opened 2019-07-11
Modernize code

chrisblossom:
This PR tackles item number 5 in [#197](https://github.com/davidtheclark/cosmiconfig/pull/197#issuecomment-504700070).

This is a fairly big PR and is not necessarily meant all to be merged. It was easier for me to do this all at once and keep the commits separated (pretty easy to follow this way) to see what actually changed. I expect there are things in here that you don't want / or want me change differently.

FYI, one goal with [`use async await where possible`](https://github.com/davidtheclark/cosmiconfig/pull/205/commits/99338dca1929740b3a7c8bf938ce94268323e5f5) was to make the `sync` and `async` functions look as similar as possible.


test/failed-directories.test.ts
```diff
@@ -1,6 +1,13 @@
...
 import { TempDir } from './util';
-import { cosmiconfig, Options } from '../src';
+import { cosmiconfig as cosmiconfigActual, Options } from '../src';
+
+const cosmiconfig: typeof cosmiconfigActual = (...params) =>
+  require('../src').cosmiconfig(...params);
+
+cosmiconfig.loadJs = cosmiconfigActual.loadJs;
+cosmiconfig.loadJson = cosmiconfigActual.loadJson;
+cosmiconfig.loadYaml = cosmiconfigActual.loadYaml;
```
davidtheclark: Can you say more about the value added by these dynamic require changes?

src/createExplorer.ts
```diff
@@ -80,83 +80,90 @@ class Explorer {
...
-              return prevResult;
-            }
-            return this.loadSearchPlace(dir, place);
-          },
-        );
-      },
-      Promise.resolve(null),
-    );
+    for await (const place of this.config.searchPlaces) {
+      const tryToLoadConfig = await this.loadSearchPlace(dir, place);
```
davidtheclark: How about naming this variable `placeResult` instead? The fact that `tryToLoadConfig` starts with a verb makes it seem like a function name.

chrisblossom:
> Man, you can really see the virtue `async/await` in making code _look_ as simple as it conceptually _is_ — whereas raw Promises can make it look more complex.

Definitely agree. Especially with `async/await` supporting both `sync` and `async` functions you don't need to add the additional `return Promise.resolve().then...` format.

> It doesn't look like you pushed each commit individually, so we don't have a CI pass on every commit. Did you test each commit locally before committing? If so that would help confirm that simultaneous source and test changes did not result in unexpected behavior changes.

I think I did, but i just ran `rm -rf node_modules && yarn && npm run check:all` on each individual commit and they all passed.

test/failed-directories.test.ts
chrisblossom: Using dynamic require is the only way I know how to get jest mocks/spys to correctly work with `util.promisify` (among other issues). Without the change the tests fail.

https://github.com/jest-community/eslint-plugin-jest/issues/283

src/createExplorer.ts
chrisblossom: Updated.

test/failed-directories.test.ts
davidtheclark: Were you facing any problems in this project, that needed to be remedied? If not, I'd prefer not to add the complexity of this commit.

chrisblossom: Are you referring to the complexity of adding:

```js
cosmiconfig.loadJs = cosmiconfigActual.loadJs;
cosmiconfig.loadJson = cosmiconfigActual.loadJson;
cosmiconfig.loadYaml = cosmiconfigActual.loadYaml;
```
or
`import { cosmiconfig as cosmiconfigActual, Options } from '../src';`
or 
`const cosmiconfig: typeof cosmiconfigActual = (...params) =>
  require('../src').cosmiconfig(...params);`

Or all three? I ask because I'm planning on putting a PR that exports the loaders as a named export `defaultLoaders` will remove the first example.

In the meantime I'll try and find an alternate solution. I think the issue comes from using `util.promisify` so I can just manually make `fs.readFile` a promise again.

chrisblossom: I reverted the commit and changed:

```js
import { promisify } from 'util';

const fsReadFileAsync = promisify(fs.readFile);
```

to: 

```js
async function fsReadFileAsync(
  pathname: string,
  encoding: string,
): Promise<string> {
  return new Promise((resolve, reject): void => {
    fs.readFile(pathname, encoding, (error, contents): void => {
      if (error) {
        reject(error);
        return;
      }

      resolve(contents);
    });
  });
}
```

PR #207, opened 2019-07-16
Prevent pushing and publishing with unformatted files

chrisblossom:
Adds [`prettier --check`](https://prettier.io/docs/en/cli.html#check) that will exit 1 if all files are not correctly formatted.


davidtheclark:
👍

PR #211, opened 2019-09-05
Separate async sync api

chrisblossom:
This PR separates the `async` and `sync` apis from each other as described in #209.

```js
import { cosmiconfig, cosmiconfigSync } from 'cosmiconfig';

const explorer = cosmiconfig({
	loaders: {
		'.js': async () => {},
		'.ts': () => {}, // can be sync
	},
	transform: async (config) => config,
	// transform: (config) => config, // can also be sync
});

await explorer.search();
await explorer.load();

const explorerSync = cosmiconfigSync({
	loaders: {
		'.js': () => {},
		'.ts': async () => {}, // can not be async, invalid
	},
	transform: (config) => config,
	// transform: async (config) => config, // can not be async, invalid
});

explorerSync.search();
explorerSync.load();
```

I think the readme could be updated further to be more condensed and clear by adding a note about the `cosmiconfigSync` separation and removing all references to what was previously `loadSync` and `searchSync`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,9 @@ If you are still using v4, those v4 docs are available [in the `4.0.0` tag](http
   - [moduleName](#modulename)
 - [explorer.search()](#explorersearch)
   - [searchFrom](#searchfrom)
-- [explorer.searchSync()](#explorersearchsync)
+- [explorerSync.search()](#explorersyncsearch)
 - [explorer.load()](#explorerload)
-- [explorer.loadSync()](#explorerloadsync)
+- [explorerSync.load()](#explorersyncload)
 - [explorer.clearLoadCache()](#explorerclearloadcache)
 - [explorer.clearSearchCache()](#explorerclearsearchcache)
 - [explorer.clearCaches()](#explorerclearcaches)
@@ -70,7 +70,7 @@ Tested in Node 8+.
 Create a Cosmiconfig explorer, then either `search` for or directly `load` a configuration file.
 
 ```js
-const { cosmiconfig } = require('cosmiconfig');
+const { cosmiconfig, cosmiconfigSync } = require('cosmiconfig');
 // ...
 const explorer = cosmiconfig(moduleName);
 
@@ -92,8 +92,10 @@ explorer.search()
 explorer.load(pathToConfig).then(..);
 
 // You can also search and load synchronously.
-const searchedFor = explorer.searchSync();
-const loaded = explorer.loadSync(pathToConfig);
+const explorerSync = cosmiconfigSync(moduleName);
+
+const searchedFor = explorerSync.search();
+const loaded = explorerSync.load(pathToConfig);
 ```
 
 ## Result
@@ -131,7 +133,7 @@ explorer.search([searchFrom]).then(result => {..})
 
 Searches for a configuration file. Returns a Promise that resolves with a [result] or with `null`, if no configuration file is found.
 
-You can do the same thing synchronously with [`searchSync()`].
+You can do the same thing synchronously with [`cosmiconfigSync.search()`].
 
 Let's say your module name is `goldengrahams` so you initialized with `const explorer = cosmiconfig('goldengrahams');`.
 Here's how your default [`search()`] will work:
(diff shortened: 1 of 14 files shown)


davidtheclark:
Nice work, @chrisblossom. I've been on vacation. I'll try to review this within the next few days so we can keep moving towards v6.

src/createExplorerBase.ts
```diff
@@ -0,0 +1,146 @@
...
+import {
+  CosmiconfigResult,
+  ExplorerOptions,
+  ExplorerOptionsSync,
+  Cache,
+  LoadedFileContent,
+} from './types';
+import { Loader } from './index';
+
+class ExplorerBase<T extends ExplorerOptions | ExplorerOptionsSync> {
```
davidtheclark: Curious: why is this `T extends A | B` rather than `A | B`?

src/createExplorerBase.ts
```diff
@@ -0,0 +1,146 @@
...
+function nextDirUp(dir: string): string {
+  return path.dirname(dir);
+}
+
+function getExtensionDescription(filepath: string): string {
+  const ext = path.extname(filepath);
+  return ext ? `extension "${ext}"` : 'files without extensions';
+}
+
+export { ExplorerBase, getExtensionDescription };
```
davidtheclark: Should we change this filename to `ExplorerBase` since it's no longer exporting a factory function?

test/index.test.ts
```diff
@@ -1,14 +1,17 @@
...
+import * as createExplorerSync from '../src/createExplorerSync';
+import { loaders } from '../src/loaders';
 
-// @ts-ignore
-const createExplorerMock: typeof createExplorer & jest.Mock = createExplorer;
+let createExplorerMock = jest.spyOn(createExplorer, 'createExplorer');
+let createExplorerSyncMock = jest.spyOn(
+  createExplorerSync,
+  'createExplorerSync',
+);
```
davidtheclark: Does `jest.mock` not work to accomplish the equivalent?

davidtheclark: Do we need to `spyOn` here if we're doing it in the `beforeEach` hook?

README.md
```diff
@@ -364,9 +366,7 @@ Do whatever you need to, and return either a configuration object or `null` (or,
 `null` indicates that no real configuration was found and the search should continue.
 
 It's easiest if you make your custom loader function synchronous.
-Then it can be used regardless of whether you end up calling [`search()`] or [`searchSync()`], [`load()`] or [`loadSync()`].
-If you want or need to provide an async-only loader, you can do so by making the value of `loaders` an object with an `async` property whose value is the async loader.
-You can also add a `sync` property to designate a sync loader, if you want to use both async and sync search and load functions.
+Then it can be used regardless of whether you end up calling [`search()`], [`load()`].
```
davidtheclark: This is no longer accurate, right?

src/createExplorerBase.ts
chrisblossom: Generics are used here because `ExplorerOptions` are not compatible with `ExplorerOptionsSync` so you have to explicitly tell typescript what object it is expecting so it can return the correct type. At least that is how I understand it.

If you try removing the generic and replacing it with `ExplorerOptions | ExplorerOptionsSync` You'll see several errors in `createExplorerSync.ts`.

src/createExplorerBase.ts
chrisblossom: Sounds good to me. I think we should also rename the class files and move the factory creation into `index.ts`. How does that sound?

test/index.test.ts
chrisblossom: `jest.mock` removes the implementation of the file, while `jest.spyOn` does not (by default) which then tests the application instead of the mock we created.

chrisblossom: For testing, no. I added it so typescript can infer the correct type. Maybe there is a better way to do it, but I am unsure how?

src/createExplorerBase.ts
davidtheclark: Sure.

chrisblossom:
Sorry for the delay, been very busy lately but I should have had this done already. I am unsure if you're wanting me to push directly to this branch, let me know if you don't want me to.

I think I addressed all issues we discussed in #211.

I do not particularly like how I fixed the spy instances in `index.test.ts`. I also ran into a strange jest issue where `toMatchObject` was not working with `transform: expect.any(function)`. If someone has a better solution I'm all ears. I also think they could be completely removed as they don't really test anything important.

PR #214, opened 2019-09-16
Load js-yaml package lazily

vankop:
**Performance fix**

This change speeds up loading `cosmiconfig` by 25-30ms. Related to https://github.com/stylelint/stylelint/issues/2454

Without this changes running [this script](https://gist.github.com/vankop/f83bdeda2cae617b4a801d8e210454ef#file-timing-js)
` node ../timing.js ./cosmiconfig/dist/index.js | head -20` results in loading time:
```
[
    [
        "./cosmiconfig/dist/index.js",
        37.836043
    ],
    [
        "./createExplorer",
        35.80891
    ],
    [
        "js-yaml",
        24.679122
    ],
    [
        "./lib/js-yaml.js",
        24.380547
    ],
    [
        "./js-yaml/loader",
        22.443439
```

With this changes applied script results for me in:
```
[
    [
        "./cosmiconfig/dist/index.js",
        7.645415
    ],
    [
        "./createExplorer",
        6.135605
    ],
    [
        "parse-json",
        1.828721
    ],
    [
        "./getPropertyByPath",
        1.266719
    ],
    [
        "import-fresh",
        1.037137
```


davidtheclark:
@vankop We're about to incorporate this change into V6. But I'm wondering: how does it affect *execution* time? Does the faster initial loading come at the expense of slower first-execution time? If so, I'd be interested in hearing your reasoning for why that's preferable.

vankop:
> But I'm wondering: how does it affect _execution_ time? Does the faster initial loading come at the expense of slower first-execution time? If so, I'd be interested in hearing your reasoning for why that's preferable.

Let's create some examples with some constants:
1. Loading `cosmiconfig` with lazy `js-yaml` takes 7ms
2. Loading `js-yaml` takes 25ms
3. Code execution takes `N` ms

1st sample - just `require('cosmiconfig')` takes 7ms
2nd sample - require takes 7ms + first-execution without `js-yaml` takes `N`ms
3d sample - require takes 7ms + first-execution with `js-yaml` takes 25ms + `N`ms

Now it works constantly:
require takes (7 + 25)ms  + first-execution takes `N`ms

So answering on your questions:
- No, it does not affect execution time
- Yes, it affects execution with `js-yaml`

However, **all** execution time is the same or faster (without `js-yaml`) if require `cosmiconfig` only when needed.

eklingen:
I'm all for this change. In trying to speed up my build chain, I've noticed cosmiconfig loads js-yaml even when none of the configuration files are actually in the yaml format, so the (relatively heavyweight) dependency is being loaded for no reason. And every little bit helps.

davidtheclark:
@vankop @eklingen Yes, the YAML-less user saves time, but for everybody else the absolute time spent doesn't change. However, absolute time is probably not the most important thing here. As I understand it, the reason that almost all Node code imports dependencies initially, rather than on-demand (besides code clarity), is to *favor fast execution at the expense of initialization time* (see [this Stack Overflow thread](https://stackoverflow.com/questions/9132772/lazy-loading-in-node-js), for example). For asynchronous API users, in particular, the consequence of this change is not just that `explorer.search()` would be unexpectedly slow, but any other code would be blocked, as well, while we lazy-load a dependency.

Is the only reason to make this change to benefit those who don't use YAML? If so, I wonder if we might consider an alternative method: by analyzing the options at initialization time, we could decide which loaders to selectively import (again, favoring slower initialization over slower execution). If the change really is to favor everybody, I'd be interested in learning more about why it's important to reduce initialization time *at the expense of* execution time?

cc @chrisblossom

davidtheclark:
I prototyped what another pattern might look like in https://github.com/davidtheclark/cosmiconfig/commit/d45b1e8bc4f7de54235d6f7e9a0973920cbe4ed8. That approach would displace the loading time from execution of `explorer.search()` to execution of `cosmiconfig()` — which may or may not be preferable. I think there's no way to *selectively* load dependencies and prevent that loading from blocking execution.

I'm still not sure about this change, and will delay the v6 release briefly to decide whether we should include it or not. If it's still questionable within a day or two, might release v6 without it and we could consider adding it back in a future release.

eklingen:
For async API users: does `await`ing a `require` call also blocks other code from running? If not, that might be a way to go?

But yeah, I understand where you're coming from. In principle, I agree with you. Both about blocking I/O at initialization vs runtime, and code clarity.

On the other hand, there's a bit of a practical balancing act. How large is the chance that a dependency is actually used during the runtime of the program, and how heavy is the dependency in question to load? And how heavy is too heavy? I feel this is not an exact science. However, I think a filetype loader is one of the special cases, where unless the program walks that exact path, the dependency will never be needed. And since js-yaml isn't super lightweight, it might just be worth the effort.

Also interesting might be to think about how a particular package is used. In this case, cosmiconfig is an awesome package, used by quite a few high profile tools to load their configuration file(s). Any extra time spent on initialization here, will reflect on the packages that require this module. And this will in turn reflect on the packages that require those packages, and so on.

vankop:
> Also interesting might be to think about how a particular package is used. In this case, cosmiconfig is an awesome package, used by quite a few high profile tools to load their configuration file(s). Any extra time spent on initialization here, will reflect on the packages that require this module. And this will in turn reflect on the packages that require those packages, and so on.

In my opinion it used mostly for CLI tools, so it does not have second run, that's why I have written:
`all execution time is the same or faster (without js-yaml)`

chrisblossom:
I think the best thing to do here is revert https://github.com/davidtheclark/cosmiconfig/pull/216/commits/08ebcd5fb544ac4e232ca854922428aba7a06a14 and only require default loaders that are being used. If you want to lazy load all loaders, you could simply lazy load `cosmiconfig` yourself, or if you just want to lazy load `yaml`, you could override the default yaml parser with your own lazy-loaded parser.

I believe that is the idea behind https://github.com/davidtheclark/cosmiconfig/commit/d45b1e8bc4f7de54235d6f7e9a0973920cbe4ed8? I put something together when coming to this same conclusion, [`chrisblossom/cosmiconfig/commits/lazy`](https://github.com/chrisblossom/cosmiconfig/commits/lazy). If moving forward with either branch, we'll have to come to a solution to access the default loaders, which probably is best just pointing people directly to `dist/PATH_TO_DEFAULT_LOADER.js`.

While looking into this, I found a bug causing `cosmiconfig` [to potentially use the incorrect json loader for `package.json`](https://github.com/chrisblossom/cosmiconfig/commit/448a9760766e82cfa9bf3d6e726dc60f35a21996#diff-d3ca038af452a97ba1e1dbf562c295b1L90).

davidtheclark:
Ok, I don't know — this morning I'm feeling like we should go for it (the change already in v6) and just see if it bothers anybody. If it turns out to be a problem, we can revert it.

> While looking into this, I found a bug causing cosmiconfig to potentially use the incorrect json loader for package.json.

I don't think this is a bug. A standard JSON loader is not the equivalent of a `package.json`-loader; and the way to customize `package.json` loading is with `packageProp`.

Issue #206, opened 2019-07-11
Remove yarn.lock


chrisblossom:
I would like to drop `yarn.lock` from this repo. Lock files are not a good fit for libraries because they are ignored by the end user consuming the library.

Basically our tests can pass, but will be broken for users. (which also why it is a good idea to _"[Reinstalls dependencies to ensure your project works with the latest dependency tree](https://github.com/sindresorhus/np#why)"_ on publish.)

This is especially true for yarn because yarn requires all people developing the software to use the same version of yarn, which is not currently enforced in this library. The `yarn.lock` format can (and has) changed between minor yarn updates.

I've also noticed that the newer versions of yarn defaults to exact versions, so we now have [a mix of exact and ^ inside package.json](https://github.com/davidtheclark/cosmiconfig/blob/47d8951caf1ed5730e948d0aa0c65309eb53208f/package.json#L108-L142) which should be fixed.

https://www.twilio.com/blog/lockfiles-nodejs

davidtheclark:
This was done in #216.

PR #216, opened 2019-10-06
V6

davidtheclark:
Development of V6 has taken longer than expected, so I've opened the branch `v6` for that work and reverted master to 5.2.1 to avoid confusion.

We'll merge v6 work into this branch, then merge it into master when it's ready for release.


chrisblossom:
Sorry for the delay, been very busy lately but I should have had this done already. I am unsure if you're wanting me to push directly to this branch, let me know if you don't want me to.

I think I addressed all issues we discussed in #211.

I do not particularly like how I fixed the spy instances in index.test.ts. I also ran into a strange jest issue where toMatchObject was not working with transform: expect.any(function). If someone has a better solution I'm all ears. I also think they could be completely removed as they don't really test anything important.

davidtheclark:
@chrisblossom No problem! I agree that you addressed everything in #211, so merged it. In https://github.com/davidtheclark/cosmiconfig/pull/217 I made a change to `index.test.ts` that I think cleans it up a little — best I could think of while maintaining the type validation.

Are there any other outstanding issues for v6 that anybody knows about? @chrisblossom @olsonpm @sudo-suhas?

sudo-suhas:
Perhaps a migration guide could be added? I realise that changelog serves this purpose to some extent but the significant changes warrant a dedicated document IMO.

chrisblossom:
#212 is still open. Should I merge those changes into this branch?

> #206: I've also noticed that the newer versions of yarn defaults to exact versions, so we now have [a mix of exact and ^ inside package.json](https://github.com/davidtheclark/cosmiconfig/blob/b89752e3120e9761ca7321ddd9f235669db39634/package.json#L113-L149) which should be fixed.

The last thing I'd like to do right before release is do a final package update and fix what is described above. Let me know when I should do this.

> In #217 I made a change to index.test.ts that I think cleans it up a little — best I could think of while maintaining the type validation.

The mock resets were added to ensure each test does not affect the other. I think I added `jest.afterEach -> jest.restoreAllMocks()` in-order to solve https://github.com/facebook/jest/issues/6059, but it looks like tests are passing without them. If something comes up later, this is most likely the cause.

Also in regards to the changes I made in `index.test.ts`, I used `toMatchInlineSnapshot`, but jest does not support having multiple tests call a single snapshot. I got around it by commenting out one of the tests before the snapshot was written, then un-commenting. This is a bit of a hack and might cause issues down the road.

> Perhaps a migration guide could be added? I realise that changelog serves this purpose to some extent but the significant changes warrant a dedicated document IMO.

Is there a good example of a migration guide that comes to mind? I don't have time right now to write a in-detailed guide, but I could do something simple. Although I think how I [documented changes to the changelog](https://github.com/davidtheclark/cosmiconfig/blob/v6/CHANGELOG.md) it serves as a mini-migration guide already.

olsonpm:
I haven't had much time recently but changing js-yaml to yaml is the only thing that comes to mind (to avoid potential future security issues).  I can probably get around to changing that dependency this weekend.

davidtheclark:
> #212 is still open. Should I merge those changes into this branch?

Ah, yeah: let's rebase the relevant commits from that and reassess, @chrisblossom.

> The last thing I'd like to do right before release is do a final package update and fix what is described above. 

Ok: I have no problem with using exact versions. Is that what you had in mind, @chrisblossom?

I believe that manually calling `jest.restoreAllMocks()` was unnecessary because we have `"restoreMocks": true` in the Jest config in `package.json`. If you set that to `false`, you'll see tests fail.

> Also in regards to the changes I made in index.test.ts, I used toMatchInlineSnapshot, but jest does not support having multiple tests call a single snapshot. I got around it by commenting out one of the tests before the snapshot was written, then un-commenting. This is a bit of a hack and might cause issues down the road.

Interesting, @chrisblossom ... let's reassess that and see if we can think of something else. I'll try to take a look soon.

I think the changelog are sufficient for migration. Although it's a breaking change, it's not a very tough one to adapt to.

> I haven't had much time recently but changing js-yaml to yaml is the only thing that comes to mind (to avoid potential future security issues).

@olsonpm sounds great 👍

olsonpm:
well I'm sorry, but this is the second time I've attempted to do something rather simple and end up spending most of my time trying to figure out how to work with types.  I was using typescript at my previous job but that was 6 months ago and so I haven't had to touch them since.  The last time I encountered this on cosmiconfig was with flow which I had never used.  The whole type ecosystem is more frustrating than it is helpful for me in javascript and it's hard for me to justify my time spent working with them.

Anyway I'm going to stop helping out with this project.  I did enjoy working with you guys - you're very smart and easy to communicate with which I appreciated a lot.

chrisblossom:
> well I'm sorry, but this is the second time I've attempted to do something rather simple and end up spending most of my time trying to figure out how to work with types. I was using typescript at my previous job but that was 6 months ago and so I haven't had to touch them since. The last time I encountered this on cosmiconfig was with flow which I had never used. The whole type ecosystem is more frustrating than it is helpful for me in javascript and it's hard for me to justify my time spent working with them.
> 
> Anyway I'm going to stop helping out with this project. I did enjoy working with you guys - you're very smart and easy to communicate with which I appreciated a lot.

I'm sorry to hear that. I've had similar issues in the past with other projects and can be very frustrating. I hope it wasn't caused by something I changed in this PR.

[Is this the package you were trying to include?
](https://www.npmjs.com/package/yaml) If so, what issues were you having? Looks like there is an [issue with their typescript definition files](https://github.com/eemeli/yaml/issues/102).

> Ok: I have no problem with using exact versions. Is that what you had in mind, @chrisblossom?

I do not think we should use exact versions and should be using `^`.

> #212 is still open. Should I merge those changes into this branch?
> 
> Ah, yeah: let's rebase the relevant commits from that and reassess, @chrisblossom.

I'll try and get this done in the next couple days.

> I believe that manually calling jest.restoreAllMocks() was unnecessary because we have "restoreMocks": true in the Jest config in package.json.

This is exactly the issue described in https://github.com/facebook/jest/issues/6059 that does not work as expected.

olsonpm:
nah, it wasn't you.  I didn't detail my issue because types will be ongoing in this repo so I thought it wasn't relevant.

For sake of you guys picking up where I left off however, I wanted to add [the prettyErrors yaml option](https://eemeli.org/yaml/#options) which wasn't in the type definition files.  The correct approach then I thought was to add our own local types until that definition file is updated, but I couldn't figure out how to merge our local type changes with upstream's.  I remembered solving that issue back at my last job and I remembered the solution being unintuititive which is when I stopped.

chrisblossom:
I think the best solution is to wait for their types to be fixed/send a PR to fix their types.

In the meantime we should set the `yaml` package as type `any`. This can be done by adding the file `global.d.ts` at the root of the project with the line:
`declare module 'yaml';`

I can pick this back up. Should I work on what you've already done or do you think it is best for me to start fresh?

olsonpm:
You can start fresh, the amount of code is so small haha.

chrisblossom:
I think I solved all open issues including removing the hacky code from `index.test.ts`. I'm okay with merging and releasing at this PR's current state as long as everyone else is.

@olsonpm can you confirm this is what you had in mind: https://github.com/davidtheclark/cosmiconfig/pull/216/commits/60337c9e1604d4d919f71edfc87baac677b83df9

Questionable commits:

#### `Object.freeze(defaultLoaders)`: https://github.com/davidtheclark/cosmiconfig/pull/216/commits/7a0f0e60954a3f2768f7f74e465c376ef0280434

I am not sure if `Object.freeze` this will slowdown the initial load time of this library. I doubt it will considering how small of an object it is.

#### Remove `yarn.lock` https://github.com/davidtheclark/cosmiconfig/pull/216/commits/fad0ebcd7be9c27a2747fef4908ab26c31767845

This closes #206. If this is not wanted, please revert the commit and close the issue. It isn't a big deal to me either way. I added it because it was the last thing I could think of that I might have put a PR in for `v6`.

olsonpm:
i'm not at my computer but the only thing i would add is the file name in the error message similar to how the error is caught and the filename prepended to error message in the loader above

chrisblossom:
> i'm not at my computer but the only thing i would add is the file name in the error message similar to how the error is caught and the filename prepended to error message in the loader above

Good idea. I'll add that tomorrow and hopefully figure out why the windows tests are failing.

test/caches.test.ts
```diff
@@ -520,7 +520,9 @@ describe('with cache disabled, does not cache file results', () => {
   });
 });
 
-describe('ensure import-fresh is called when loading a js file', () => {
+// This test does not work with lazy loading JS -- TODO: fix or remove
+// eslint-disable-next-line jest/no-disabled-tests
+describe.skip('ensure import-fresh is called when loading a js file', () => {
```
chrisblossom: I am unsure how to fix this test. I don't understand the point of the test, don't we know that by requiring it?. I personally think we should just remove it and move on.

chrisblossom:
I took the idea from #214 and lazy loaded all default loaders. Any negatives from doing this?

davidtheclark:
@chrisblossom Great work.
- `Object.freeze` seems fine to me.
- The lazy loading seems like a good idea and I don't know any reasons not to. We'll see if anybody finds any!
- I don't know if I'm fully on board about `yarn.lock` but I or others can re-add it later if we think it's important, no big deal.
- Deleting that import-fresh test seems ok to me. I will say after this experience I'm kind of skeptical about whether it's worth type-checking test files, because it seems to make stubbing/mocking/spying so much harder, with unclear benefits. But we have it working for now!

I pushed commit ed83eaa72ab4ecbf4ce3dc115d62341f3a74dee4 with some documentation and changelog adjustments. I think we're ready to launch.

Can anybody please chime in if you think we need to wait? If not, I'll merge & publish the release tomorrow (Sunday).

chrisblossom:
I can't think of anymore changes and I'm good with merging / releasing.

> Deleting that import-fresh test seems ok to me. I will say after this experience I'm kind of skeptical about whether it's worth type-checking test files, because it seems to make stubbing/mocking/spying so much harder, with unclear benefits. But we have it working for now!

The `import-fresh` issue had nothing to do with types. It is a jest mocking / require issue.

davidtheclark:
I released V6 this morning — decided not to spent more time fussing about the lazy requiring, and just see if users have problems with it (the stakes, I think, are pretty low). Nice work! 🍰 🍰 🍰

PR #225, opened 2019-12-31
Fix a link

KSXGitHub:
Before this change, `cosmiconfig()` in ToC points to the very first heading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ If you are still using v5, those v5 docs are available [in the `5.x.x` tagged co
 - [Usage](#usage)
 - [Result](#result)
 - [Asynchronous API](#asynchronous-api)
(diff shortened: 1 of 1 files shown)


davidtheclark:
Thanks, @KSXGitHub. The trouble with this is that the table of contents is automatically generated, by `remark-preset-davidtheclark` which uses [`remark-toc`](https://github.com/remarkjs/remark-toc). I'll file an issue there to see what's up.

davidtheclark:
I think I found the bug and submitted a PR to fix: https://github.com/syntax-tree/mdast-util-toc/pull/64

davidtheclark:
Replaced by https://github.com/davidtheclark/cosmiconfig/pull/230.

PR #229, opened 2020-01-15
move @types dependency to devDependencies

antonk52:
Hi, looks like the users of cosmiconfig had to download `@types/parse-json` when adding cosmiconfig to their project. From what I saw this dependency is only needed when working on cosmiconfig directly. Therefore, it makes sense to keep it in devDependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,6 @@
     ]
   },
   "dependencies": {
-    "@types/parse-json": "^4.0.0",
     "import-fresh": "^3.1.0",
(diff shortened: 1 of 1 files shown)


davidtheclark:
Looks to me like this is pretty standard practice with Typescript. For a long discussion, see https://github.com/microsoft/types-publisher/issues/81. In that thread, the reasoning to keep the `@types` in `dependencies` seems convincing to me.

PR #232, opened 2020-02-25
Break on duplicate config

deanshub:
It's useful to us to break on multiple configurations being defined,
so I added a flag for this.


davidtheclark:
Never got CI to pass on this PR.

I'm not inclined to add this feature. I think it's possible in userland (start a new search after you find a config). Unless there are more requests for similar functionality, I'll leave it in userland.

PR #243, opened 2020-11-28
feat(loaders): adds the ability to use merge (`<<`) syntax

jun-sheaf:
The `yaml` package has turned `merge` syntax off in recent versions. This PR adds `merge` syntax back in.


davidtheclark:
Why was this syntax removed without a major version change? Is it unsafe? Is it nonstandard? Can you please provide more context about the situation?

Issue #224, opened 2019-11-25
support `mjs` extensions out of box


alexander-akait:
If you approve this i can send a PR

davidtheclark:
@evilebottnawi Yeah, I'd be curious what implementation you think will work best. Is there a third-party loader that does this well already? 

Looks like maybe the [woefully underdocumented `import()` function](https://nodejs.org/api/esm.html#esm_code_import_code_expressions) might work for `.mjs`, wrapped in some error handlers for incompatible Node versions?

xiaoxiangmoe:
```js
function importConfig(configSrc){
   return (process.versions.node greaterOrEqual '13.2.0') ? import(configSrc) : Promise.resolve(require(configSrc));
}
```

evilebottnawi:
@davidtheclark what do you think above example above?

davidtheclark:
Why do we need to check the Node version instead of `typeof import !== 'undefined'`? 

Does `import()` work with CommonJS files (`.cjs` or often `.js`)? The docs don't explain that, or the details about `x.default`. Is this better documented somewhere else? I'm hesitant to rely on precedent from Babel or Webpack or Rollup. But we should have tests verifying that it works, anyway.

Relying on `import()` is going to cause some problems with the synchronous API. Is there a synchronous alternative?

evilebottnawi:
Only `import` without `then`, i will try to find time for implementing PoC and we will continue the discussion :+1: Busy right now, sorry, if somebody want to go ahead, PR welcome

evilebottnawi:
Anyway i think for `cjs` it should be easy

xiaoxiangmoe:
@davidtheclark import is not a function，so we can't use typeof import. import cjs will export module {default: cjsModule}

k2snowman69:
Just curious, what is the status on this? I see there hasn't been an update in here for a few months now. Also, could we consider splitting cjs and mjs into two separate PRs? 

If I'm not mistaken, currently this is preventing the use of the `"type": "module"` flag in a package.json for more recent versions of node (12.17.0 and higher) and at minimum supporting the extension `.cjs` with the current code should unblock the use of the `"type": "module"` property.

https://nodejs.org/api/esm.html#esm_package_json_type_field

k2snowman69:
Just in case the answer is yes to the cjs support, I opened #238 to speed things along

coreyfarrell:
[...]

PR #251, opened 2021-02-13
Support ECMAScript modules (ESM)

davidtheclark:
This PR intends to close https://github.com/davidtheclark/cosmiconfig/issues/224. I don't have more time to work on this today so I'll get what I have up as a draft — feedback welcome! **I'd especially appreciate help testing this from people interested in using ESM,** both ensuring that it works as expected and ensuring that the expectations are the right ones.

Some notes:

- ESM is only supported by the asynchronous API, because it is enabled by the asynchronous `import()` function.
- So `.mjs` is only added to the default `searchPlaces` of the asynchronous API.
- I turned off the selective branches in CI configuration because I wasn't seeing tests run. It may be that Travis and Appveyor needed to catch up with the main branch name change (from `master` to `main`). We'll see.
- I was having some trouble early on with Jest, so tried using the `jest-circus` test runner. Turns out that wasn't the problem at all, but I left in that update.
- This PR is set to merge into a `v8` branch, in case there's anything else in the issue queue that anybody wants to get into v8.

### DIFF
--- a/src/loaders.ts
+++ b/src/loaders.ts
@@ -3,17 +3,30 @@
 import parseJsonType from 'parse-json';
 import yamlType from 'yaml';
 import importFreshType from 'import-fresh';
-import { LoaderSync } from './index';
-import { LoadersSync } from './types';
+import { LoaderSync, LoaderAsync } from './index';
+import { Loaders } from './types';
+import { canUseDynamicImport } from './canUseDynamicImport';
 
 let importFresh: typeof importFreshType;
-const loadJs: LoaderSync = function loadJs(filepath) {
+const loadJsSync: LoaderSync = function loadJsSync(filepath) {
   if (importFresh === undefined) {
     importFresh = require('import-fresh');
   }
 
   const result = importFresh(filepath);
   return result;
+};
+
+const loadJs: LoaderAsync = async function loadJs(filepath) {
+  // This logic is validated by tests running in Node versions that don't
+  // support dynamic imports, like Node 10.
+  /* istanbul ignore else */
+  if (canUseDynamicImport()) {
+    const imported = await import(filepath);
+    return imported.default;
+  } else {
+    return loadJsSync(filepath, null);
+  }
 };
 
 let parseJson: typeof parseJsonType;
@@ -46,6 +59,6 @@ const loadYaml: LoaderSync = function loadYaml(filepath, content) {
   }
 };
 
-const loaders: LoadersSync = { loadJs, loadJson, loadYaml };
+const loaders: Loaders = { loadJs, loadJsSync, loadJson, loadYaml };
 
 export { loaders };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -112,6 +112,7 @@ module.exports = {
     // Many tests make assertions indirectly in a way the plugin
     // does not understand.
     'jest/expect-expect': 'off',
+    'jest/no-identical-title': 'off',
   },
   settings: {
     node: {
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,15 +3,15 @@ git:
   depth: 5
 cache: yarn
 branches:
-  only:
-    - main
-    - v7
+  # only:
+  #   - main
+  #   - v7
 matrix:
   include:
     - node_js: '10'
-      after_success: 'bash <(curl -s https://codecov.io/bash)'
     - node_js: '12'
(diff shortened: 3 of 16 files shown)


davidtheclark:
Known to dos:

- [ ] Document change to `defaultLoaders` (`loadJs` is async) and exposure of `defaultLoadersSync`.
- [ ] Write new tests that find ESM configs.

davidtheclark:
By using the `--experimental-vm-modules` flag and some workarounds for the inability to clear the ESM loader cache, the tests are now passing for me locally. But on Travis, they are not, with segmentation faults in Node 10 and 14 and something else in Node 12 I don't understand yet.

src/loaders.ts
```diff
@@ -16,6 +17,18 @@ const loadJs: LoaderSync = function loadJs(filepath) {
   return result;
 };
 
+const loadJs: LoaderAsync = async function loadJs(filepath) {
+  // This logic is validated by tests running in Node versions that don't
+  // support dynamic imports, like Node 10.
```
glen-84, line 22: Is it specific versions of Node 10 that don't support dynamic import syntax?

Using [this site](https://www.jdoodle.com/execute-nodejs-online/), 9.2.0 gives a SyntaxError, but 10.1.0 doesn't.

I think this will fail when the syntax is supported, but the feature is not (f.e. when `--experimental-modules` is needed).

Since v10 EOL is just around the corner, it's probably easier to just wait until then (2021-04-30), and drop support for it.

sindresorhus: > Since v10 EOL is just around the corner, it's probably easier to just wait until then (2021-04-30), and drop support for it.

I would just drop support for Node.js 10 now. It's 10 days left. This will not be released until then anyway.

src/loaders.ts
```diff
@@ -16,6 +17,18 @@ const loadJs: LoaderSync = function loadJs(filepath) {
...
+  /* istanbul ignore else */
+  if (canUseDynamicImport()) {
+    const imported = await import(filepath);
```
lohfu, line 25: An import call will syntax error in node versions that do not support it. In a [PR](https://github.com/substack/tape/pull/547) I'm fiddling with to support ESM in [tape](https://github.com/substack/tape) I side step this syntax error by conditionally requiring another file containing the import statement...

The conditional require:
https://github.com/lohfu/tape/blob/esm-support/bin/tape#L29

The separate file:
https://github.com/lohfu/tape/blob/esm-support/bin/load-module-esm-support.js

Since dynamic imports seem to be up to 5 times slower than require, we conditionally use import or require by checking file extensions and [get-package-type](https://github.com/cfware/get-package-type). As you probably load far less files this might not be necessary for you.

lohfu: If you dont want to conditionally use require vs import, you could just use the function you create to check dynamic import support, https://github.com/davidtheclark/cosmiconfig/pull/251/files#diff-e61e569af4ed5c2bd19c648e11225dd502296f0931b339c88cb82288a493e6adR6.

ie something like:

```ts
let dynamicImport;

try {
  dynamicImport = new Function('id', 'return import(id);');
} catch (e) {} 

// later

const loadJs: LoaderAsync = async function loadJs(filepath) {
  if (dynamicImport()) {
    const imported = await dynamicImport(filepath);

    return imported.default;
  } else {
    return loadJsSync(filepath, null);
  }
};
```

lohfu: it appears it's even trickier... versions 10.0.0 - 12.16 and 13.0 - 13.1 support the `import` keyword, but will throw a "Not Supported" error when called. See https://github.com/inspect-js/has-dynamic-import/issues/2 for an attempted work around.

perrin4869:
The segfault issue is probably caused by https://github.com/nodejs/node/issues/35889

davidtheclark:
Just posted this note at the top of the README:

> MAINTAINERS WANTED! If you're interested in taking over and maintaining Cosmiconfig, please let @davidtheclark know (with an issue or email). I'd like to hand over the keys completely, so I'm looking for owners, not people who just want to merge a PR or two! You can make the decisions about what happens in v8 and subsequent versions, how the package balances stability and opinionated features, and so on. Take a look at open issues and PRs to learn about possibilities that have been on people's minds over the years.

If you're interested in taking ownership of the package and moving this feature forward, let me know!

d-fischer:
This was superseded by #304, which is now merged.

PR #260, opened 2021-08-21
Do not crash if there is a directory named after a search place

tjenkinson:
Recently fixed a very similar bug in eslint 😆 

refs https://github.com/rollup/plugins/pull/981

### DIFF
--- a/test/successful-directories.test.ts
+++ b/test/successful-directories.test.ts
@@ -780,6 +780,52 @@ describe('searchPlaces can include subdirectories', () => {
   });
 });
 
+describe('directories are not treated as files', () => {
+  beforeEach(() => {
+    temp.createFile('a/.foorc.json', '{ "found": true }');
+    temp.createDir('a/b/package.json/c');
+  });
+
+  const startDir = temp.absolutePath('a/b/package.json/c');
+  const explorerOptions = {
+    stopDir: temp.absolutePath('.'),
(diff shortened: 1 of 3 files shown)


test/successful-directories.test.ts
```diff
@@ -780,6 +780,52 @@ describe('searchPlaces can include subdirectories', () => {
   });
 });
 
+describe('directories are not treated as files', () => {
```
davidtheclark, line 783: I think a more verbose description or multi-sentence comment would be helpful for explaining this test case. It takes some very close reading to realize that there's a directory named `package.json`, and that's the focus of the test. Could we call out what's going on here?

davidtheclark (commented):
Ah, interesting! Thanks for the bug fix 👍 

Would you mind either adding a `## HEAD` section to `CHANGELOG.md` where you describe the change, or just writing up in a comment what you'd like to see in the changelog? Since you're familiar with the bug you probably have the best context for writing down what exactly got fixed.

test/successful-directories.test.ts
tjenkinson: Pushed a commit. Is that clearer?

davidtheclark: Yeah, looks good 🙇

davidtheclark (approved):
Great! I'll merge & release this later today.

PR #263, opened 2021-10-08
Add .config subdir to default searchPlaces

### LINKED ISSUE #262: Standardize a common configuration folder in the defaults
As pointed out in https://github.com/davidtheclark/cosmiconfig/issues/152#issuecomment-479111308:

> It would be great to have a solution like OP pointed out, or as others have suggested: <project_root>/.config

> That means cosmiconfig has a golden opportunity to be a leader in standardizing / consolidating schemes.

> The core problem for me is that my projects are cluttered with config files at the top level and I'd like to have the ability to personally wrangle that.

> **End users don't have much control over where their config files are located**

To sum it up, I think it would be great to *default* to additionally searching in a predefined subdirectory (I think `.config` seems like the most popular one).

The benefit of this would be that you wouldn't have to clutter your package.json nor your root folder with tooling configuration, both of which seem to be annoying for many people. Rather than this being up to the library/tooling author that uses `cosmicconfig`, this should be up to the end user using said library/tooling (see last quote), which is why (IMO) appealing to library/tooling authors to customize `searchPlaces` is not the right approach.

You'd still be able to do either of these two ways; this is really just adding another option by default.

Do note that this is not a duplicate of #152 since the XDG config spec is for **user-global** configuration while this aims to standardize a common **per-project** configuration folder.

~~I am open to create a PR for this.~~ EDIT: I already created #263 :)

d-fischer:
fixes #262

Open questions:

- do we want to add the `.config.ext` variants inside the config folder? (yes/no)
- do we want to keep the `rc` suffix? (yes/no/only for the extensionless file/add both variants/...?)

Also there should probably be adjustments to the tests and the README :)


kentcdodds:
How can we help get this over the finish line?

d-fischer:
> How can we help get this over the finish line?

I don't know - the maintainer has shown no interest in this at all.

kentcdodds:
Looks like @davidtheclark is no longer very active in OSS and the last commit was by @Codex-. Any chance this could get in Alex? 🙏

kentcdodds:
Oh, my bad, just realized Alex's commit was merged by @davidtheclark. I noticed that @sudo-suhas has publish rights on the package though. Any chance you could help us here?

kentcdodds:
Maybe switch to lilconfig? https://github.com/antonk52/lilconfig/issues/27

d-fischer:
I mean, this issue/PR proposes a solution where tooling maintainers don't have to do anything to fix this problem (unless they have non-default config in some specific keys).

Proposing to switch libraries is a thing you'd have to do with every single tooling maintainer, and if they don't like it, we're back to square one.

kentcdodds:
Agreed, it would be better to have this added to cosmiconfig, but if no maintainer is maintaining it then that won't happen. Just suggesting an alternative. I've definitely personally been the maintainer of a library I was tired of maintaining and was thrilled to deprecate it in favor of something else.

MichaelDeBoey:
As I said on Twitter: it would be an extra hurdle to convince all these packages to switch to `lilconfig`, especially since it would be a breaking change because `lilconfig` handles files without an extension differently
https://github.com/antonk52/lilconfig#difference-to-cosmiconfig

d-fischer:
The problem is that this is a breaking change to _all users of the tool_. If a YAML loader is not explicitly given, all existing YAML config files will break.

Anyway, I don't think a "call to replace this" is on topic here.

d-fischer:
I just published this PR as a fork!

To use it as an end user, you need to use yarn resolutions or npm overrides.

## yarn

```json
  "resolutions": {
    "cosmiconfig": "npm:@d-fischer/cosmiconfig@7.1.0"
  }
```

## npm

```json
  "overrides": {
    "cosmiconfig@7": "npm:@d-fischer/cosmiconfig@7.1.0"
  }
```

Then just move your config files to a `.config` subfolder :)

# Caveats
- obvoiusly, this only works with tools that actually use `cosmiconfig@^7.0.0` or higher. Some tools may be behind, some other tools (such as eslint) may not use `cosmiconfig` at all - if you want to move these files, they need to update their dependencies or change their other/custom solution!
- Even worse, with yarn, this might actually break packages that use a lower major version. YMMV!
- If the tool developer customized the configuration (specifically, the `searchPlaces` option) this doesn't work either. Again, I can't do anything about this, please show them that you want this!

davidtheclark:
Just posted this note at the top of the README:

> MAINTAINERS WANTED! If you're interested in taking over and maintaining Cosmiconfig, please let @davidtheclark know (with an issue or email). I'd like to hand over the keys completely, so I'm looking for owners, not people who just want to merge a PR or two! You can make the decisions about what happens in v8 and subsequent versions, how the package balances stability and opinionated features, and so on. Take a look at open issues and PRs to learn about possibilities that have been on people's minds over the years.

If you're interested in taking ownership of the package and moving this feature forward, let me know!

MichaelDeBoey:
@davidtheclark If nobody wants to take over ownership, you can maybe contact @openjs-foundation to ask if they want to take over the project.

d-fischer:
If there are no hard objections, I will merge this into a new branch that will branch off main **before the js-yaml change** and try to manage getting a minor version out that includes this.

PR #270, opened 2022-05-27
Switch from yaml to js-yaml

### LINKED ISSUE #266: Update yaml dependency
cosmiconfig depends on `yaml@1.x` which leads to the following warning in the console of projects that depend on something using cosmiconfig.

```
(node:3672) [DEP0148] DeprecationWarning: Use of deprecated folder mapping "./" in the "exports" field module resolution of the package at .../node_modules/yaml/package.json.
```

This was fixed in `yaml@2.x`.

davidtheclark:
Closes #266.

See some background discussion in https://github.com/stylelint/stylelint/issues/6101, too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "import-fresh": "^3.2.1",
     "parse-json": "^5.0.0",
     "path-type": "^4.0.0",
-    "yaml": "^1.10.0"
+    "js-yaml": "^4.1.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.10.4",


davidtheclark:
Looks like Travis testing isn't working any more. Opened https://github.com/davidtheclark/cosmiconfig/pull/271 to switch to GitHub Actions.

jakebailey:
Shouldn't this also contain a change to the imports to specify js-yaml instead? This might only be working because something else in node_modules depends on the yaml package and therefore the import still worked.

davidtheclark:
@jakebailey YES. Good call. I wasn't thinking well this morning. Open to a PR, or I will make sure to fix before releasing.

PR #274, opened 2022-05-27
Prepare v8

davidtheclark:
I'll leave this open for a day to see if any watchers have concerns.


chrisblossom:
Might want to drop node 12 as well because it is [https://github.com/nodejs/Release](no longer supported). Also should add the upcoming LTS v18 to the test matrix.

davidtheclark:
I don't want to drop support for Node 12 because at least one key user, Stylelint, still supports Node 12 and we don't have a reason of our own to drop support.

We have v18 in the test matrix -- did you have something else in mind? I'm open to whatever.

JounQin:
> The [js-yaml](https://github.com/nodeca/js-yaml) package doesn't strictly follow the YAML 1.2 spec. It supports some YAML 1.1 features.

https://github.com/unifiedjs/unified-engine/issues/65

---

Should that be considered? @davidtheclark

chrisblossom:
> I don't want to drop support for Node 12 because at least one key user, Stylelint, still supports Node 12 and we don't have a reason of our own to drop support.

Because this is a breaking change with the new YAML parser- stylelint would need to bump to a new major to use it, right? I assume stylelint would drop unsupported node versions at the same time as a version bump. 

> We have v18 in the test matrix -- did you have something else in mind? I'm open to whatever.

Sorry, very limited cell phone service for the week so I missed that v18 was added.

If you’d like, I could work on updating dependencies before the version bump if that would be helpful. I would need a week or so though to find time.

ybiquitous:
The Stylelint's current test suite passes the next version 8 of `cosmiconfig` (this PR's branch). 🎉 
See https://github.com/stylelint/stylelint/pull/6120 

---

> > I don't want to drop support for Node 12 because at least one key user, Stylelint, still supports Node 12 and we don't have a reason of our own to drop support.
> 
> Because this is a breaking change with the new YAML parser- stylelint would need to bump to a new major to use it, right? I assume stylelint would drop unsupported node versions at the same time as a version bump.

I think it's possible that Stylelint's next major version will drop the Node 12 support (due to EOL), but the Stylelint team doesn't discuss anything about it yet.

davidtheclark:
Asking Stylelint about both yaml/js-yaml and Node version support over in https://github.com/stylelint/stylelint/issues/6101#issuecomment-1140465191.

ybiquitous:
Hi @d-fischer, the Stylelint team has been going forward to the next major v15.0.0. 

It would be nice to release Stylelint v15, including a dependency on Cosmiconfig v8. I will be happy to help if there is anything I can do.

ybiquitous:
Thanks, I understand the situation and agree with your plan. I will review the PRs.

PR #275, opened 2022-05-27
Fix js-yaml, type imports

jakebailey:
This fixes the use of `js-yaml` by `npm i`-ing to fix the lock, adding the right types package, and then calling the right function.

I also made a minor loader fix; the imports at the top to get the type for the lazily-loaded modules will actually cause an import to occur immediately. I can revert this part if you'd like.

There are some other type errors to do with the thrown errors, but they don't appear when running `tsc` because the version of TypeScript this repo uses is super out of date (far past what DefinitelyTyped supports, so I would strongly suggest upgrading).

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -1,1 +1,0 @@
-declare module 'yaml';
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,11 +9,10 @@
       "version": "7.0.1",
       "license": "MIT",
       "dependencies": {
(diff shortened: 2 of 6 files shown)


davidtheclark (approved):
Thanks for fixing my facepalmworthy mistake! When CI passes, we're good to merge, I believe.

SimenB:
Will this get a release, or is this blocked by waiting for a new maintainer (#277)?

PR #280, opened 2022-11-01
Add support for `*.mjs`, `*.config.mjs`

JulianCataldo:
Hello there!

It would be useful to support `*.mjs`, what do you people think?

It originated from here:
https://github.com/mdx-js/eslint-mdx/issues/436#issuecomment-1296084347

Cheers


d-fischer:
This won't work, as ES modules are loaded differently from CJS modules (`.js` and `.cjs`). #251 is already a much more intricate effort to support this. It probably needs to be properly rebased to the current state of the code, and the open to-dos resolved. Feel free to expand upon that PR, as its original author is now relatively inactive in this repository.

PR #288, opened 2023-01-07
Give back some power to the end users

### LINKED ISSUE #261: cosmiconfig-(meta)-config
Hey! Great library!
Using it for a few projects of mine, and that is a problem - the more I use it - the more different files in different places I do have.

## Proposal
- Allow cosmiconfig configuration via config files in the same way it powers end-solutions
- Allow tapping into config file parsing and discovery, so I can configure other solutions using single configuration file. This is similar, if not equal to having config in `package.json`, just not bound to the package boundary.
```
// metaconfig.yml
lintstyled:
  lintstyled-config
prettier:
  prettier-config
```
- Allow programmatic override of the functionality above, so I can configure the fleet of my tools in a different way using a "virtual file system", so it would be completely transparent to the end users.
```
// cosmiconfig.js
export const tapVirtualFile = (folder, pattern) => {
 if (specialFolders.include(folder)) {
   return specialConfigs[pattern]
 }
}
```

----

There is no real need to implement this, but I have to describe the opportunity and the possible use cases.

d-fischer:
This change intends to make configuration more configurable for the end users, to suit the end user's project structure rather than the structures that (possibly multiple different) tooling maintainers imagined.

## Examples

With a file named `.config.yml` in your project root with the following contents:

```yml
cosmiconfig:
  searchPlaces:
    - .config/{name}.yml
```

and given that you're using stylelint and prettier (with both using a version of cosmiconfig that supports this), cosmiconfig will *only* look at the file `.config/stylelint.yml` and `.config/prettier.yml`.

The same should work from a package.json "cosmiconfig" key.

An additional feature I plan to add here is to be able to put configuration for cosmiconfig consumers directly into this file, a little bit like it works now in package.json (#261):

```yml
cosmiconfig:
  searchPlaces:
    - .config/{name}.yml
  searchInThisFile: true # bikeshedding proposals encouraged

prettier:
  printWidth: 80
```

This was in a previous local version of mine, but I decided to remove it, because the general approach seemed too shaky to me, but I intend to try again.

## To do

- [x] Enable user to specify explorer options within `package.json` using a `cosmiconfig` key
- [x] Enable user to specify explorer options in a `.config.{yml,yaml,json,js,cjs}` file using a `cosmiconfig` key
- [x] Enable user to put *all* configuration in said file
- [ ] Optional: figure out a way to specify loaders from non-JS files (could also offload this to a new issue to implement later, but then we should probably throw for now when someone tries to specify the `loaders` key in a non-JS file)
  - #290 
- [x] Fix existing tests to not spy on reading this early stage
- [x] Create new tests (get coverage back to 100%)
- [x] Add documentation to README
- [x] Add short summary to CHANGELOG

Fixes #261 

This is intended to be in a *minor* update, so there should be no breaking changes.

Please add your thoughts to this, I appreciate any feedback 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Tested in Node 14+.
 
 ## Usage for tooling developers
 
-*If you are an end user (i.e. a user of a tool that uses cosmiconfig),
+*If you are an end user (i.e. a user of a tool that uses cosmiconfig, like `prettier` or `stylelint`),
 you can skip down to [the end user section](#usage-for-end-users).*
 
 Create a Cosmiconfig explorer, then either `search` for or directly `load` a configuration file.
--- a/README.md
+++ b/README.md
@@ -607,8 +607,7 @@ The following property is currently actively supported in these places:
 
 ```yaml
 cosmiconfig:
-  # overrides where configuration files are being searched to enforce a common naming convention and format
-  # (which are defined by yourself)
+  # overrides where configuration files are being searched to enforce a custom naming convention and format
   searchPlaces:
     - .config/{name}.yml
 ```
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Cosmiconfig continues to search up the directory tree, checking each of these pl
 ## Table of contents
 
 - [Installation](#installation)
-- [Usage](#usage)
+- [Usage for tooling developers](#usage-for-tooling-developers)
 - [Result](#result)
 - [Asynchronous API](#asynchronous-api)
   - [cosmiconfig()](#cosmiconfig-1)
@@ -56,6 +56,7 @@ Cosmiconfig continues to search up the directory tree, checking each of these pl
   - [ignoreEmptySearchPlaces](#ignoreemptysearchplaces)
 - [Caching](#caching)
 - [Differences from rc](#differences-from-rc)
+- [Usage for end users](#usage-for-end-users)
 - [Contributing & Development](#contributing--development)
 
 ## Installation
@@ -66,7 +67,10 @@ npm install cosmiconfig
 
 Tested in Node 14+.
 
-## Usage
+## Usage for tooling developers
+
+*If you are an end user (i.e. a user of a tool that uses cosmiconfig),
+you can skip down to [the end user section](#usage-for-end-users).*
 
 Create a Cosmiconfig explorer, then either `search` for or directly `load` a configuration file.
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -539,6 +543,96 @@ To avoid or work around caching, you can do the following:
...
+  `.${moduleName}rc.cjs`,
+  `.config/${moduleName}rc`,
+  `.config/${moduleName}rc.json`,
+  `.config/${moduleName}rc.yaml`,
+  `.config/${moduleName}rc.yml`,
+  `.config/${moduleName}rc.js`,
+  `.config/${moduleName}rc.cjs`,
+  `${moduleName}.config.js`,
+  `${moduleName}.config.cjs`,
+]
```
tjjfvi, line 570: Perhaps this would be clearer if in the same format as the `searchPlaces` config:
```
package.json
.{name}rc
.{name}rc.json
...
```

README.md
```diff
@@ -539,6 +543,96 @@ To avoid or work around caching, you can do the following:
...
+```
+
+The following properties are currently actively supported in these places:
+
+```yaml
+cosmiconfig:
+  # overrides where configuration files are being searched, this might be a minor performance optimization
+  # if all your configuration files use the same naming convention and format
+  searchPlaces:
+    - .config/{name}.yml
```
tjjfvi, line 615: This isn't one of the formats listed above; was it supposed to be `.config/{name}rc.yml`, or is this a separate search location?

README.md
```diff
@@ -539,6 +543,96 @@ To avoid or work around caching, you can do the following:
...
+      
+  # enables configuration resolution in this same file, similarly to `package.json`, see below
+  searchInThisFile: true
```
tjjfvi, line 618: Any reason for this to not always be on?

test/index.test.ts
```diff
@@ -1,28 +1,23 @@
...
-    {},
-  );
-
-  return loaderFunctionsByName;
+  return Object.entries(loaders).reduce((acc, [extension, loader]) => {
+    return {
+      ...acc,
+      [extension]: loader.name,
+    };
+  }, {});
```
tjjfvi, line 20:
```suggestion
  return Object.fromEntries(Object.entries(loaders).map(([extension, loader]) => [extension, loader.name]));
```

test/index.test.ts
```diff
@@ -171,10 +170,12 @@ describe('cosmiconfig', () => {
       },
     };
 
-    const checkResult = (mock: jest.SpyInstance) => {
-      expect(mock.mock.calls.length).toEqual(1);
-      expect(mock.mock.calls[0].length).toEqual(1);
-      const { transform, loaders, ...explorerOptions } = mock.mock.calls[0][0];
+    const checkResult = (mock: jest.SpyInstance, instanceNum: number) => {
```
tjjfvi, line 173: The two `checkResult` functions in this file are largely identical except for different option objects; possibly worth extracting to a parameter and sharing the logic.

README.md
d-fischer: This is supposed to show that you can do whatever you want here.

README.md
d-fischer: Good point, I think I had a reason, but I forgot. Will re-check.

README.md
d-fischer: `searchPlaces` actually uses the same format as this, but I made it a bit clearer anyway since this is end user documentation.

README.md
d-fischer: I see now - the reason why this exists is because it will conflict with the `ignoreEmptySearchPlaces` configuration that tool authors can set. If that is set, and searching in the meta config file is enabled (i.e. `searchInThisFile`), the explorer will not look anywhere else and return an empty configuration. I will add a note for this.

d-fischer: Actually, I guess I'll fix that instead.

README.md
tjjfvi: Ah, in that case it might be worthwhile to consider replacing the blurb about "minor performance optimization". I don't think end users will really care about a ms-level startup improvement.

README.md
tjjfvi: Ah, so that means that if you add
```yaml
# .config.yml
cosmiconfig:
  searchInThisFile: true
```
It may cause some subset of tools using cosmiconfig to ignore previously-working config files? That seems non-ideal. I'd only expect behavior like that on something like
```yaml

# .config.yml
cosmiconfig:
  searchInThisFile: true
  
prettier:
```

d-fischer: I fixed this issue and removed the `searchInThisFile` flag.

README.md
```diff
@@ -539,6 +543,90 @@ To avoid or work around caching, you can do the following:
...
+.config.js
+.config.cjs
+```
+
+The following property is currently actively supported in these places:
+
+```yaml
+cosmiconfig:
+  # overrides where configuration files are being searched to enforce a common naming convention and format
+  # (which are defined by yourself)
```
tjjfvi, line 611:
```suggestion
  # overrides where configuration files are being searched to enforce a custom naming convention and format
```

tjjfvi (approved):

fisker:
This approach seems bad to me.

1. It seems we can't config cosmiconfig for different tools.

	 For example, I want Prettier look for `prettier.config.mjs` and stylelint look for `stylelint.config.cjs` it seems can't be done.

2. The new behavior seems not able to opt-out.

	I don't want to allow user override searchPlaces.

Maybe I was wrong?

d-fischer:
> 1. It seems we can't config cosmiconfig for different tools.
> 
> 	 For example, I want Prettier look for `prettier.config.mjs` and stylelint look for `stylelint.config.cjs`, can it be done? At least, not clear to me how to do it.

Just configure both extensions. `searchPlaces` is still an array as usual. (I thought the placeholder was described clearly enough.)

> 2. The new behavior seems not able to opt-out.
> 
> 	I don't want to allow user override searchPlaces.

Why would you deliberately want to say that your library/tool doesn't want to adhere to the end user's naming conventions? This is a strict improvement for end users and I don't see why consumers would have to have any say in this.

d-fischer:
Maybe the first point wasn't actually made clear enough in the library. The `{name}` in the example isn't some kind of documentation placeholder, but it's a real "variable" that will be replaced by the consumer's name (the first argument to `cosmiconfig()`)  at runtime.

fisker:
> Why would you deliberately want to say that your library/tool doesn't want to adhere to the end user's naming conventions? 

If Prettier want support this, shouldn't this enough?

```js
const {config: {searchPlaces}} = cosmiconfig("cosmiconfig").search()
const {config: prettierConfig} = cosmiconfig("prettier", {searchPlaces}).search()
```

No one requested it, why would we make things complicate?

d-fischer:
> > Why would you deliberately want to say that your library/tool doesn't want to adhere to the end user's naming conventions? 
> 
> If Prettier want support this, shouldn't this enough?
> 
> ```js
> const {config: {searchPlaces}} = cosmiconfig("cosmiconfig").search()
> const {config: prettierConfig} = cosmiconfig("prettier", {searchPlaces}).search()
> ```

So you propose that *every single tool maintainer* wanting to support this needs to copy and paste a snippet of code into their own code base, which pretends they are cosmiconfig? That's the whole point of a library completely taken apart. After a while, there would be subtle incompatibilities, not to mention that many (like you, apparently) don't even realize the problem that this solves, and in turn, probably won't include it. After all, no one made this before now, and no one requested it either, which is the perfect segue to your next question:

> No one requested it, why would we make things complicate?

The problem is that tooling maintainers are usually living inside their own bubble. Yes, no tooling maintainer may have requested this, but I've heard a bunch of voices saying that their application, which uses a bunch of tools that have config files, all put these config files into the root of the project, which becomes annoying quickly. The addition of the `.config` subdirectory into the default `searchPlaces` was the first step towards improving this, but more steps are necessary for a complete solution, and this isn't even the last one. After all, not everyone uses `cosmiconfig` for config loading, and most people who just use e.g. `prettier` aren't even aware of cosmiconfig's existence.

Now, after I drifted away from the question a bit, I shall answer with a question of my own: **Why are you so strictly against a new, optional feature that tooling maintainers don't even really have to care about?**

If you don't want people to use it, just don't tell them. For the most part, at least at first, people will still be unaware of this feature because they're unaware of cosmiconfig itself even though it's in their dependency tree, possibly even multiple times.

fisker:
Never mind. Hope other tool maintainers like it.

Just for the record, "No one requested it" I mean I didn't see any Prettier user asked for it in Prettier repository.

PR #294, opened 2023-03-04
Migrate tests to vitest

beerose:
Related to https://github.com/cosmiconfig/cosmiconfig/pull/283
CleanShot 2023-03-14 at 15 26 22@2x


test/util.ts
```diff
@@ -1,10 +1,12 @@
 import path from 'path';
 import del from 'del';
+import * as fsType from 'fs';
 import makeDir from 'make-dir';
 import parentModule from 'parent-module';
 import os from 'os';
+import { Mock, SpyInstance, vi } from 'vitest';
 
-const fs = jest.requireActual('fs');
+const fs = await vi.importActual<typeof fsType>('fs');
```
nix6839, line 9:
```suggestion
const fs = await vi.importActual<typeof import('fs')>('fs');
```

darkobits:
Thanks for this, @beerose! 🙌 

You should consider adding the following to conserve the coverage thresholds that [were in place via `package.json` previously](https://github.com/cosmiconfig/cosmiconfig/pull/294/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L85-L92):

> `vite.config.ts`

```ts
{
  test: {
    coverage: {
      branches: 100,
      functions: 100,
      lines: 100,
      statements: 100
    }
  }
}
```

May need to either reduce those figures slightly or get those few stay lines covered to keep the 100% targets.

beerose:
@darkobits done. I added the coverage config to `vite.config.ts` and used `istanbul` provider (as before). The coverage is now at 100%.

.eslintrc.js
```diff
@@ -20,11 +20,10 @@ module.exports = {
     'plugin:import/errors',
     'plugin:import/warnings',
     'plugin:import/typescript',
-    'plugin:jest/recommended',
     'prettier',
     'prettier/@typescript-eslint',
   ],
-  plugins: ['jest', '@typescript-eslint', 'import'],
+  plugins: ['@typescript-eslint', 'import'],
```
d-fischer, line 26: Are there equivalent rules in `eslint-plugin-vitest`? I have not checked in depth, but might be worth looking at later.

src/index.ts
```diff
@@ -13,18 +13,17 @@ import {
 } from './types';
 
 type LoaderResult = Config | null;
+export type LoaderSync = (filepath: string, content: string) => LoaderResult;
```
d-fischer, line 16: This reordering seems unnecessary, please revert to the original order

d-fischer (changes requested):

src/index.ts
beerose: Done. I originally did that because it broke `no-use-before-define` rule, but since you have it disabled for `@typescript/no-use-before-defined`, and disabled that one too.

.eslintrc.js
beerose: There are equivalents for most — I added it in my latest commits. The only ones that I didn't find equivalents for are: `prefer-todo` (but there are no usages of `test.todo`) and `require-to-throw-message` (disabled anyway).

beerose: Yeah, sorry, pushed now.

d-fischer (approved):

d-fischer:
Thanks for your effort - now other things can move on 🎉

PR #304, opened 2023-04-28
Support loading ES Modules

beerose:
Continuation of https://github.com/cosmiconfig/cosmiconfig/pull/283


src/index.ts
```diff
@@ -86,6 +81,7 @@ const defaultLoaders = Object.freeze({
   noExt: loaders.loadYaml,
 } as const);
 const defaultLoadersSync = Object.freeze({
+  '.mjs': loaders.loadJsSync,
```
d-fischer, line 84: I don't think that works. ESM can only load async.

src/index.ts
```diff
@@ -195,11 +185,15 @@ function cosmiconfigSync(
   };
 }
 
+function normalizeOptions(
+  moduleName: string,
+  options: Options,
+): ExplorerOptions;
```
d-fischer: This overload seems useless, since it has the same signature as the implementation.

d-fischer (changes requested):

src/index.ts
beerose: Yes, it's still a draft. Working on it

tianyingchun:
it seems that this change aim to `native esm` with `.mjs` 

but while our `pckage.json` declare`type:module` with source file extention `.ts`,`.js`, it will as `esm` also.

beerose:
> it seems that this change aim to `native esm` with `.mjs`
> 
> but while our `pckage.json` declare`type:module` with source file extention `.ts`,`.js`, it will as `esm` also.

AFAIK that's something that typescript loader will take care of. @d-fischer can you confirm?

src/index.ts
beerose: Yep, made a mistake when solving conflicts. Updated now

package.json
```diff
@@ -92,10 +89,10 @@
...
   },
   "devDependencies": {
-    "@babel/cli": "^7.10.4",
-    "@babel/core": "^7.10.4",
-    "@babel/preset-env": "^7.10.4",
-    "@babel/preset-typescript": "^7.10.4",
+    "@babel/cli": "7.19.3",
+    "@babel/core": "7.20.2",
+    "@babel/preset-env": "7.20.2",
+    "@babel/preset-typescript": "7.18.6",
```
d-fischer, line 95: Any reason why this switches to exact versioning?

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 8.1.4
+
+- **Breaking change:** Add support for ECMAScript modules (ESM) to the [*asynchronous* API](./README.md#asynchronous-api). End users running Node versions that support ESM can provide `.mjs` files, or `.js` files whose nearest parent `package.json` file contains `"type": "module"`.
```
d-fischer, line 5: Is this really a breaking change?

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 8.1.4
```
d-fischer, line 3: Should be added as `HEAD`. Versioning concerns are not PR concerns.

package.json
```diff
@@ -25,12 +25,6 @@
     "check:all": "npm run test && npm run typescript && npm run lint && npm run format:check",
     "prepublishOnly": "npm run check:all && npm run build"
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "lint-staged && npm run typescript && npm run test",
-      "pre-push": "npm run check:all"
-    }
-  },
```
d-fischer, line 33: Where did these go?

d-fischer (changes requested):

src/index.ts
```diff
@@ -91,33 +101,25 @@ function replaceMetaPlaceholders(
...
+  options: T,
+): T {
   const metaExplorer = new ExplorerSync({
     packageProp: 'cosmiconfig',
     stopDir: process.cwd(),
     searchPlaces: metaSearchPlaces,
     ignoreEmptySearchPlaces: false,
     usePackagePropInConfigFiles: true,
-    loaders: defaultLoaders,
+    loaders: { ...defaultLoadersSync, ...defaultLoaders },
```
d-fischer, line 114: We can't really add async loaders here since it's an inherently sync explorer.

package.json
beerose: I think they were removed in the previous esm-support PR (the one I'm building on top of). Will add it back

package.json
beerose: I suppose not.

CHANGELOG.md
beerose: Can you elaborate? What should I change it to?

d-fischer: It should literally just be `HEAD`, as previous PRs have done it as well. Assigning a real version to this is part of the release procedure.

d-fischer: No you did not get it. This is not part of 8.1.3, which is already released. The convention is to literally write "HEAD" as the version.

beerose: Done.

d-fischer (approved):

tianyingchun:
it seems has a bug here 
```ts
const loadJs: LoaderAsync = async function loadJs(filepath) {
  try {
    const { href } = pathToFileURL(filepath);
    return (await import(href)).default;
  } catch (error) {
    return loadJsSync(filepath, null);
  }
};
```
if we have named export here it will always return `undefined`

maybe it should be changed to 
```ts
const { href } = pathToFileURL(filepath);
const moduleData =  await import(href);
return moduleData.default.|| moduleData;
```

d-fischer:
I don't see why it should load a non-default export from the file. Do you have any actual use case for this? I'm pretty sure that vite does not use cosmiconfig.

PR #311, opened 2023-07-31
feat: add support for TypeScript configuration files

jrandolf:
This PR implements first-party support for TypeScript configuration files.

The implementation is designed to be minimal. The TypeScript configuration will be transpiled to CommonJS or ESM, depending on which API is used. The feature target is set to ES2020 since this version is fully compatible with Node.js 16, the last maintained version of Node.js.

TypeScript is left to be a peer dependency with optional installation, so this doesn't bulk-up downstream downloads.


ghiscoding:
this PR seems to be causing some problems when trying to update cosmiconfig to v8.3.x in Lerna-Lite, I got a few failing unit tests as shown in this Renovate [Lerna-Lite PR](https://github.com/lerna-lite/lerna-lite/pull/713). Are there anything I need to change on my side in order to support this? cc @d-fischer

d-fischer:
@ghiscoding It seems like when there is no file, it does not return null anymore in your case. I have tried to construct a failing test against this, but I always get `null` where I should be. Could you please assist with figuring out specifics and then create a new issue for this?

ghiscoding:
@d-fischer the behavior seem to have changed and I opened issue #325 for this.

Issue #333, opened 2023-10-15
Offer a "minimal" version with less dependencies


d-fischer:
There should be a more minimal (or "base") package offered that gets rid of a few dependencies:

- `parse-json` could be replaced by `JSON.parse` - the former improves error messages, but if you don't want that, the latter reduces the dependency impact to zero since it's included in the standard library
- `import-fresh` could be replaced by just calling `require` - similar to above, the advantage of the former being that file changes during the runtime can be respected
- `js-yaml` could be removed, making YAML support optional

This would leave us with only two dependencies, `env-paths` and `path-type`, which each have no dependencies of their own. (We can probably also get rid of `path-type`, but that's a separate concern.)

The end goal would probably be that there is a "base" package, and the "full" package depends on that, adding and replacing default loaders to achieve the current state of `cosmiconfig` (i.e., this should not contain any breaking changes).

You should be able to use the "base" package and still add back a YAML parser, `parse-json` or `import-fresh` as you like.

It is important to note that it should be possible to ignore `searchPlaces` that do not have loaders attached to them due to the mentioned removal of YAML support in the "base" package. If you add a YAML parser (or the "full" package does), the `searchPlaces` should be in the exact state that is currently in the code.

oliveryasuna:
Check out [lilconfig](https://www.npmjs.com/package/lilconfig).

PR #340, opened 2024-02-23
Proposal: Monorepo

Phillip9587:
WIP: Do not review yet

#333


d-fischer:
I'm gonna stop you right there - this is splitting out way too much. Having an unhealthy amount of tiny packages is not the goal of the issue.

PR #342, opened 2024-02-23
Proposal: Monorepo

Phillip9587:
## This proposal exist to start the discussion on this topic. It is not a finished implementation.

I created a basic monorepo setup with 2 packages: `cosmiconfig` and `@cosmiconfig/core`. 

### Quotes from Issue #333

> There should be a more minimal (or "base") package offered that gets rid of a few dependencies:

`@cosmiconfig/core` only depends on `env-paths`. It offers the same `cosmiconfig(...)` and `cosmiconfigSync(...)` PublicExplorer methods. 

It also exports 2 new internal methods `createPublicExplorer()` and `createPublicExplorerSync()` for the creation of these methods. 

> parse-json could be replaced by JSON.parse - the former improves error messages, but if you don't want that, the latter reduces the dependency impact to zero since it's included in the standard library

I refactored the `loadJSON` loader of the core package to use `JSON.parse(...)` and removed the `parse-json` dependency. The cosmiconfig package contains the `loadJSON` loader with better errors from `parse-json`. 

> import-fresh could be replaced by just calling require - similar to above, the advantage of the former being that file changes during the runtime can be respected

I refactored the `loadJsSync` loader of the core package to use `require()` and removed the `import-fresh` dependency. The cosmiconfig package contains the `loadJsSync` loader with the `import-fresh` package. 

> js-yaml could be removed, making YAML support optional

Done. The core package has no loader for `.yml` and `.yaml`. The cosmiconfig package contains the yaml loader.

> This would leave us with only two dependencies, env-paths and path-type, which each have no dependencies of their own. (We can probably also get rid of path-type, but that's a separate concern.)

`path-type` was already removed so that leaves us with `env-paths`.

> The end goal would probably be that there is a "base" package, and the "full" package depends on that, adding and replacing default loaders to achieve the current state of cosmiconfig (i.e., this should not contain any breaking changes).

The `cosmiconfig` package depends on `@cosmiconfig/core` and adds back the "full" cosmiconfig loaders, searchPlaces, globalSearchPlaces, etc.


d-fischer:
This looks mostly fine to me, except the root README file should be updated to reflect the features opened up here. Also, the separate packages require README files as well, so the npm pages don't end up empty.

packages/cosmiconfig/test/getPropertyByPath.test.ts
```diff
@@ -0,0 +1,92 @@
+import { describe, test, expect } from 'vitest';
+import { getPropertyByPath } from '../src/util';
```
d-fischer, line 2: I don't think this will work - the file does not seem to exist. We should just be able to delete this test file- it already exists in the core package.

tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
-    "rootDir": "./src",
-    "outDir": "./dist",
-    "stripInternal": true
+    "types": ["@types/node"],
+    "noEmit": true,
+    "allowJs": true
```
d-fischer, line 6: Why was allowJs enabled?

tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
   "extends": "./tsconfig.base.json",
   "compilerOptions": {
-    "rootDir": "./src",
-    "outDir": "./dist",
-    "stripInternal": true
+    "types": ["@types/node"],
```
d-fischer, line 4: Why are these types explicitly referenced? They should be automatically loaded because they're global.

d-fischer (changes requested):

Rchie121 (approved):

Rchie121 (approved):

Issue #351, opened 2025-09-08
Supply-chain attack cause by parse-json > error-ex module


ACHP:
The last version of the ([github.com/qix-/node-error-ex](https://github.com/qix-/node-error-ex)) module include malicious code as reported https://github.com/Qix-/node-error-ex/issues/17.

This module is not a direct dependency of cosmiconfig but a dependency of [parse-json ](https://github.com/sindresorhus/parse-json/) version < 8.0.0. ( Nov 2, 2023 )

PR #352, opened 2025-09-16
fix(security): remove parse-json to address security vulnerability

agustints:
## Overview

This update addresses a critical security vulnerability (CVSS 9.3) in the `error-ex` package, a transitive dependency of `parse-json`. This vulnerability poses a significant risk, especially for applications executed in browser-like environments that interact with cryptocurrency wallets.

### Affected Package

- **Package**: `error-ex@1.3.4`
- **Vulnerability ID**: SNYK-JS-ERROREX-12554213
- **Severity**: Critical (CVSS 9.3)
- **CWE**: CWE-506 (Embedded Malicious Code)

### Dependency Chain (historical)

```
cosmiconfig@8.3.6 → parse-json@5.2.0 → error-ex@1.3.4
```

### Malicious Behavior

The `error-ex` package has been reported to contain malicious code that:

1. Monitors network traffic when executed in a browser context
2. Targets cryptocurrency transactions
3. Hooks into Web3 wallet interactions
4. Intercepts and redirects ETH transactions to attacker-controlled addresses
5. Tamper-modifies swap and transfer transactions

### High Risk Scenarios

- Applications running in browser environments
- Projects that interact with Web3 wallets or cryptocurrency
- Any application where users may perform crypto transactions

### Potential Consequences

- **Financial loss**: Redirection of transactions to attacker wallets
- **Data exposure**: Network interception risk
- **Reputation damage**: User trust erosion due to compromises
- **Regulatory exposure**: Compliance issues following a breach

### What changed in cosmiconfig

- Replaced runtime usage of `parse-json` in the JSON loader with the built-in `JSON.parse`.
- Preserved the public error surface by wrapping the native error message to match existing expectations: `JSON Error in <filepath>:` followed by the original parse error.
- Maintained behavior for empty files and both sync/async flows; verified by running the full test suite (no regressions).
- Avoided ESM/CJS interop problems stemming from newer ESM-only `parse-json` versions.

This removes the transitive risk path via `error-ex` by eliminating the need for `parse-json` at runtime.

### Actions for consumers

- Update to the cosmiconfig version containing this change.
- If your application runs in browser-like environments or uses crypto/Web3, audit recent transactions and logs for anomalies.

### Prevention Measures

- Implement dependency scanning in CI/CD (e.g., Snyk, npm audit, Dependabot).
- Regularly review and prune dependency trees.
- Minimize optional dependencies that offer limited benefit but expand the attack surface.

## Why removing `parse-json` is safe here

- **Requirements**: Cosmiconfig only needs standards-compliant JSON parsing and to surface errors with a consistent prefix. It does not rely on `parse-json`'s `JSONError` class, code frames, or filename parameterization.
- **Behavior preserved**: Using `JSON.parse` produces the same parsing results. We catch and rethrow errors with the same leading message (`JSON Error in <filepath>:`), keeping the external behavior and test expectations intact.
- **Compatibility and security**: Removing the dependency avoids ESM-only interop issues in sync code paths and eliminates the historical transitive exposure to `error-ex`.
- **Verified by tests**: The full test suite passes unchanged, confirming functional parity.

## References

- **Snyk Advisory**: SNYK-JS-ERROREX-12554213
- **CVSS Score**: 9.3 (Critical)
- **CWE**: CWE-506 — Embedded Malicious Code (https://cwe.mitre.org/data/definitions/506.html)

## Status

⚠️ Critical — addressed by removing `parse-json` usage and relying on `JSON.parse` with preserved error messaging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,13 +71,11 @@
   "dependencies": {
     "env-paths": "^2.2.1",
     "import-fresh": "^3.3.0",
-    "js-yaml": "^4.1.0",
-    "parse-json": "^5.2.0"
+    "js-yaml": "^4.1.0"
   },
   "devDependencies": {
     "@types/js-yaml": "^4.0.5",
     "@types/node": "^14",
-    "@types/parse-json": "^4.0.0",
     "@typescript-eslint/eslint-plugin": "^6.5.0",
     "@typescript-eslint/parser": "^6.5.0",
     "@vitest/coverage-istanbul": "^0.34.3",
--- a/src/loaders.ts
+++ b/src/loaders.ts
@@ -38,14 +38,9 @@ export const loadJs: Loader = async function loadJs(filepath) {
   }
 };
 
-let parseJson: typeof import('parse-json');
 export const loadJson: LoaderSync = function loadJson(filepath, content) {
-  if (parseJson === undefined) {
-    parseJson = require('parse-json');
-  }
-
   try {
-    return parseJson(content);
+    return JSON.parse(content);
   } catch (error) {
     error.message = `JSON Error in ${filepath}:\n${error.message}`;
     throw error;


d-fischer:
More than a week ago, the offending `error-ex` version `1.3.3` has been deleted from npm. There is no vulnerability to be handled anymore. See also #351. `1.3.4` does not contain any compromised code. Even the linked Snyk ID says so - it specifically shows that only exactly `=1.3.3` is vulnerable.

`1.3.4` on the other hand is functionally equal to `1.3.2` and was only released so dependents using the `^1.3.3` version range or using `npm upgrade` have a version to resolve to.

Even if this vulnerability wouldn't exist, we use the `parse-json` to allow for better error messages that help you get behind the issue in your config file, so we do not want to remove it.

Please do more research before submitting any issues or PRs in the future in order to prevent time waste on either side.

PR #355, opened 2026-02-07
Removed Parse JSON

AlexanderKaran:
Removing `parseJSON`, which in turn removes 5 packages from the dependency tree: parse-json, @babel/code-frame, error-ex, json-parse-even-better-errors, and lines-and-columns.

1. parse-json internally calls JSON.parse as its primary path — it only adds enhanced error formatting (code frames via @babel/code-frame) when parsing fails.
2. Cosmiconfig called parseJson(content) with only one argument — it never used parse-json's reviver or filename parameters, and never referenced its JSONError type or
   codeFrame property.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,13 +71,11 @@
   "dependencies": {
     "env-paths": "^2.2.1",
     "import-fresh": "^3.3.0",
-    "js-yaml": "^4.1.0",
-    "parse-json": "^5.2.0"
+    "js-yaml": "^4.1.0"
   },
   "devDependencies": {
     "@types/js-yaml": "^4.0.5",
     "@types/node": "^14",
-    "@types/parse-json": "^4.0.0",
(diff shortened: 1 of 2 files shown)


d-fischer:
Originally, I only wanted this in the planned "minimal" version of cosmiconfig. However, [V8 seems to have improved its JSON.parse error reporting.](https://github.com/nodejs/node/issues/24361#issuecomment-1937994382) So this can land now; I still wanted to wait for a new major version though since error messages are part of the public interface.

AlexanderKaran:
> Originally, I only wanted this in the planned "minimal" version of cosmiconfig. However, [V8 seems to have improved its JSON.parse error reporting.](https://github.com/nodejs/node/issues/24361#issuecomment-1937994382) So this can land now; I still wanted to wait for a new major version though since error messages are part of the public interface.

Thank you so much for merging this.

PR #360, opened 2026-05-21
fix: avoid bundling typescript when the TS loader is unused

### LINKED ISSUE #358: Vite bundles TypeScript (~5MB) even when `.ts` loader is excluded
When bundling a project with Vite that uses cosmiconfig, the build includes the entire `typescript` package (~5MB) even though TypeScript config support is explicitly disabled.

### Real-world example

[`npm-check-updates`](https://github.com/raineorshine/npm-check-updates) recently converted to ESM in [version 21.0.0](https://github.com/raineorshine/npm-check-updates/blob/main/CHANGELOG.md#2100---2026-04-14) and started using cosmiconfig.
Even though `.ts` is excluded from `searchPlaces`, the Vite build pulls in the full `typescript` package.

### Reproduction

```javascript
// vite.config.ts
import { defineConfig } from 'vite'

export default defineConfig({
  build: {
    lib: {
      entry: './src/index.ts',
      formats: ['es', 'cjs'],
    },
  },
})
```

```javascript
// src/index.ts
import { cosmiconfig, defaultLoaders } from 'cosmiconfig'

const { '.ts': _, ...loaders } = defaultLoaders

const explorer = cosmiconfig('ncu', {
  searchPlaces: ['.ncurc', '.ncurc.json', '.ncurc.yaml', '.ncurc.yml', '.ncurc.mjs', '.ncurc.cjs', '.ncurc.js'],
  searchStrategy: 'global',
  loaders,
})
```

### Expected behavior

Since `.ts` is excluded from `searchPlaces` and destructured out of `loaders`, the `typescript` package should not be included in the bundle.

### Actual behavior

Vite's static analysis traces the `require('typescript')` inside `loadTsSync` in `cosmiconfig/dist/loaders.js` and bundles the entire `typescript` package:

```
build/chunks/typescript-XXXX.js  5,150.16 kB │ gzip: 1,169.39 kB
```

### Root cause

`defaultLoaders['.ts']` contains a dynamic `require('typescript')` inside a function body. Vite/Rollup cannot tree-shake this because:

1. `require()` is treated as potentially side-effectful
2. The function is exported from the module, so the bundler preserves it
3. The module graph includes `typescript` regardless of whether the loader is used

### Question

Is there an **official** way to configure cosmiconfig so that bundlers like Vite can exclude the TypeScript loader when it's not needed?

I have reviewed the docs and tried:

- Excluding `.ts` from `searchPlaces` — does not prevent bundling
- Building a custom `loaders` object without `.ts` — does not prevent bundling
- Looking for subpath exports (e.g., `cosmiconfig/loaders`) — does not exist

None of these prevent Vite from tracing the `require('typescript')` in the source.

### Potential solutions

1. **Restructure with subpath exports**: Move `loadTs`/`loadTsSync` to a separate subpath export (e.g., `cosmiconfig/loaders/typescript`) so users opt-in explicitly. This would allow bundlers to exclude the TypeScript loader entirely when not imported.

2. **Truly lazy dynamic import**: Replace the top-level `require('typescript')` in `loadTsSync` with a dynamic `import()` inside the function body, potentially with a webpack/vite magic comment for lazy loading that bundlers can optimize.

3. **Split `defaultLoaders`**: Provide a `defaultLoaders` export that excludes TypeScript by default, with a separate `defaultLoadersWithTypeScript` opt-in for users who need it.

Would you accept a PR for any of these approaches? I will be happy to contribute.

ATOM00blue:
## Problem
`src/loaders.ts` referenced the `typescript` package via a string literal (`require('typescript')` / `import('typescript')`) inside the TS loaders. Bundlers such as Vite/Rollup statically trace those specifiers and include the ~5MB `typescript` package in consumer bundles even when the `.ts` loader is never used.

## Fix
Reference the package name through a variable (`const typescriptPackageName = 'typescript'`) so the `require`/`import` specifier is not a static string. `typescript` is still resolved lazily at runtime when the loader actually runs, but bundlers can no longer statically trace it, keeping it out of bundles that don't use the TS loader.

## Test
Added `test/loaders.test.ts` verifying the TS loaders still load and transpile correctly.

Closes #358

### DIFF
--- a/src/loaders.ts
+++ b/src/loaders.ts
@@ -68,11 +68,19 @@ export const loadYaml: LoaderSync = function loadYaml(filepath, content) {
   }
 };
 
+/**
+ * The `typescript` package name is referenced through this variable rather than
+ * as a string literal so that bundlers (e.g. Vite/Rollup) cannot statically
+ * trace the `require`/`import` below. This keeps the ~5MB `typescript` package
+ * out of consumer bundles when the TypeScript loader is not used, while the
+ * module is still resolved lazily at runtime when the loader runs.
+ */
+const typescriptPackageName = 'typescript';
 let typescript: typeof import('typescript');
 export const loadTsSync: LoaderSync = function loadTsSync(filepath, content) {
   /* istanbul ignore next -- @preserve */
   if (typescript === undefined) {
-    typescript = require('typescript');
+    typescript = require(typescriptPackageName);
   }
   const compiledFilepath = `${filepath}.${randomUUID()}.cjs`;
   try {
@@ -99,7 +107,7 @@ export const loadTsSync: LoaderSync = function loadTsSync(filepath, content) {
 
 export const loadTs: Loader = async function loadTs(filepath, content) {
   if (typescript === undefined) {
-    typescript = (await import('typescript')).default;
+    typescript = (await import(typescriptPackageName)).default;
   }
   const compiledFilepath = `${filepath}.${randomUUID()}.mjs`;
   let transpiledContent;
--- a/test/loaders.test.ts
+++ b/test/loaders.test.ts
@@ -1,0 +1,60 @@
+import { readFileSync } from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+import { afterAll, beforeEach, describe, expect, test } from 'vitest';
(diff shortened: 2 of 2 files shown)


d-fischer:
Not necessary as v10 will load TS like JS, depending on Node's native TS stripping feature.

PR #362, opened 2026-07-20
fix: correctly decode UTF-16 (BOM-prefixed) config files

DanielC000:
## What changed

`Explorer.ts` and `ExplorerSync.ts` hardcoded UTF-8 when reading a config file (`fs.readFile(filepath, { encoding: 'utf-8' })` / `fs.readFileSync(filepath, 'utf8')`). A config file saved as UTF-16LE or UTF-16BE — for example via PowerShell's default `Out-File`/redirection encoding on Windows, a very common way Windows users create files like `.releaserc` — gets decoded as garbage instead of being parsed correctly.

This adds a `decodeFileContent` helper that sniffs the leading bytes for a UTF-16LE BOM (`0xFF 0xFE`) or UTF-16BE BOM (`0xFE 0xFF`) and decodes with the matching encoding via `TextDecoder`, before falling back to the existing UTF-8 path unchanged. Applied symmetrically to both the async and sync read paths.

## Why (originating report)

This was originally reported against semantic-release, which delegates all config loading to cosmiconfig: [semantic-release/semantic-release#1624](https://github.com/semantic-release/semantic-release/issues/1624) — ".releaserc is not correctly parsed when encoding of file is utf16le".

- @gr2m (semantic-release maintainer): *"We'd love a PR to make reading the config files work correctly in this case."*
- @travi (semantic-release maintainer), on revisiting in 2024: *"i imagine this is more related to cosmiconfig than directly about semantic-release. i would recommend starting there."*

semantic-release's `lib/get-config.js` delegates entirely to `cosmiconfig(CONFIG_NAME).search(cwd)`, so the actual bug lives here. There's no cosmiconfig issue yet for this — happy to open one if you'd prefer that first.

## Testing

- Added `test/decodeFileContent.test.ts`, unit-testing the new decode helper directly: UTF-16LE, UTF-16BE, plain UTF-8, and UTF-8-with-BOM (confirming the UTF-8 path — BOM or not — is byte-for-byte unchanged from before).
- Added UTF-16LE and UTF-16BE fixture-based end-to-end tests to `test/successful-files.test.ts` (async + sync), following the file's existing pattern.
- `npm test` passes: 267/267 tests, 100% coverage maintained.
- `npm run lint` clean.

### DIFF
--- a/test/decodeFileContent.test.ts
+++ b/test/decodeFileContent.test.ts
@@ -1,0 +1,51 @@
+import { describe, test, expect } from 'vitest';
+import { decodeFileContent } from '../src/util';
+
+// UTF-16BE has no built-in Node Buffer encoding, so derive it by
+// byte-swapping the UTF-16LE encoding of the same string.
+function utf16beBuffer(contents: string): Buffer {
+  const le = Buffer.from(contents, 'utf16le');
+  const be = Buffer.alloc(le.length);
+  for (let i = 0; i < le.length; i += 2) {
+    be[i] = le[i + 1];
+    be[i + 1] = le[i];
+  }
+  return be;
+}
+
+describe('with a UTF-16LE byte-order mark', () => {
+  test('decodes as UTF-16LE and strips the BOM', () => {
+    const bom = Buffer.from([0xff, 0xfe]);
+    const content = Buffer.from('{ "foo": true }', 'utf16le');
+    expect(decodeFileContent(Buffer.concat([bom, content]))).toBe(
+      '{ "foo": true }',
+    );
+  });
+});
+
+describe('with a UTF-16BE byte-order mark', () => {
+  test('decodes as UTF-16BE and strips the BOM', () => {
+    const bom = Buffer.from([0xfe, 0xff]);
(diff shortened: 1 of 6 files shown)


test/decodeFileContent.test.ts
```diff
@@ -0,0 +1,51 @@
...
+// byte-swapping the UTF-16LE encoding of the same string.
+function utf16beBuffer(contents: string): Buffer {
+  const le = Buffer.from(contents, 'utf16le');
+  const be = Buffer.alloc(le.length);
+  for (let i = 0; i < le.length; i += 2) {
+    be[i] = le[i + 1];
+    be[i + 1] = le[i];
+  }
+  return be;
+}
```
d-fischer: This seems identical to the function in `successful-files.ts`. Please move it to `test/util.ts`.

d-fischer (changes requested):

test/decodeFileContent.test.ts
DanielC000: moved it to test/util.ts, thanks

d-fischer (approved):

PR #366, opened 2026-08-08
fix: prevent search() deadlock when cwd is inside the global config dir [AI]

### LINKED ISSUE #365: `search()` never settles when cwd is inside the global config dir (`searchStrategy: 'global'`) [AI]
I stumbled across this behavior and was *very* confused. I handed off to Claude to diagnose and fix. The following is authored by Claude.

**Version:** 10.0.0

### Summary

With `searchStrategy: 'global'`, if the process's cwd is the global config
directory or any descendant, `explorer.search()` returns a promise that never
settles. The global config file is never read. In a CLI this usually surfaces as
the process exiting silently having done nothing, so it's easy to misdiagnose as
"the global config file is ignored".

A fix and regression tests are proposed in the accompanying PR.

### Repro

```bash
D=$(mktemp -d)
mkdir -p "$D/fakehome/.config/myapp" "$D/elsewhere"
echo '{"serverUrl":"https://found"}' > "$D/fakehome/.config/myapp/config.json"

(cd "$D" && npm init -y >/dev/null && npm install cosmiconfig >/dev/null)
cat > "$D/probe.mjs" <<'EOF'
const explorer = cosmiconfig('myapp', {
  searchStrategy: 'global',
  searchPlaces: ['package.json', '.myapprc', '.myapprc.json'],
});
const t = setTimeout(() => { console.log('TIMEOUT: search() never settled'); process.exit(3); }, 4000);
const r = await explorer.search();
clearTimeout(t);
console.log('RESULT:', r?.filepath ?? null);
EOF

# works
(echo works:; cd "$D/elsewhere"               && HOME="$D/fakehome" XDG_CONFIG_HOME= node "$D/probe.mjs")
# hangs
(echo fails:; cd "$D/fakehome/.config/myapp" && HOME="$D/fakehome" XDG_CONFIG_HOME= node "$D/probe.mjs")
```

Passing `cache: false` makes both cases pass.

### Cause

`getGlobalDirs()` yields the same path twice when the start dir is the global
config dir — once as the walk's start (`isGlobalConfig: false`) and once as the
terminal global yield (`isGlobalConfig: true`). Those are not interchangeable:
`getSearchPlacesForDir()` uses `globalConfigSearchPlaces` for the latter.

But `Explorer.search()` keys the cache on the path alone:

```js
return await emplace(this.searchCache, currentDir.path, search); // isGlobalConfig dropped
```

and `emplace` caches the pending promise. So the global-dir step hits the cache
entry created by the start-dir step — the promise whose body is still
executing — and the search transitively awaits itself.

(Descendant cwds fail the same way: the global dir is visited and cached during
the upward walk before the global step is reached.)

### Proposed fix

Include the flag in the cache key, via a small shared helper (`getSearchCacheKey`
in `src/util.ts`) used by both `Explorer.search()` and `ExplorerSync.search()`:

```ts
export function getSearchCacheKey(dir: DirToSearch): string {
  return dir.isGlobalConfig ? `global:${dir.path}` : dir.path;
}
```

```ts
return await emplace(this.searchCache, getSearchCacheKey(currentDir), search);
```

The entry-point `emplace(this.searchCache, from, search)` needs no change, since
`from` is always a non-global start. No cycle can be reintroduced — the global
dir is the generator's last yield.

`ExplorerSync` isn't susceptible to the deadlock itself (its `emplace` only
caches settled results, not in-flight work, so there's no promise to await
recursively), but it gets the same composite-key fix for consistency, so the
two implementations don't silently diverge.

### Regression tests

Added to `test/search-strategies.test.ts`, under the `global` strategy suite,
covering both `cosmiconfig` (async) and `cosmiconfigSync`, with cwd set to (a)
the global config dir and (b) a descendant of it. Caching is left at its
default (**on**) in all cases — the bug is invisible with `cache: false`. The
async cases carry an explicit per-test timeout so a regression reports as a
failing/timed-out test rather than hanging the whole suite.
</content>

thromer:
With searchStrategy: 'global' and caching enabled, getGlobalDirs() can yield the same directory twice with different isGlobalConfig flags (once during the upward walk, once as the terminal global-dir visit). The search cache keyed on path alone, so the second visit hit the in-flight promise from the first and awaited itself forever.

Fixes #365


src/util.ts
```diff
@@ -13,6 +14,17 @@ export function emplace<K, V>(map: Map<K, V>, key: K, fn: () => V): V {
...
 
+// Computes the key under which a directory's search result is cached. The
+// global config dir can coincide with a plain directory visited earlier in
+// the same search (e.g. when the search starts inside it), so the flag must
+// be part of the key or the two visits' cached results would collide.
+/**
+ * @internal
+ */
+export function getSearchCacheKey(dir: DirToSearch): string {
+  return dir.isGlobalConfig ? `global:${dir.path}` : dir.path;
```
d-fischer: This should probably add a prefix in both cases, since there are a decent number of filesystems that do allow colons in file and directory names. Adding prefixes in both cases prevents possible collisions.

d-fischer (changes requested):

src/util.ts
thromer: What if instead of `global:` we do `global\0` as the prefix, since IIUC `NUL` cannot appear in a file or directory name in either Windows or Unix ?

d-fischer: If we use a prefix (that's delimited with the same character) in both cases, there are zero considerations to be made about conflicts. I'm sure there's still some arcane filesystem where null bytes are okay in filenames, so let's be safe.

thromer: Done, PTAL. Nice catches!

d-fischer (approved):

